PR #3, opened 2021-01-21
Stateful Enum and Serde support

### LINKED ISSUE #1: Enum with Data compatability
Currently it is not possible to generate Typescript bindings for enums that contain data.

For example:

```
#[derive(TS)]
enum Foo {
    Bar(String), 
    Other    
}
```
Produces an error: "variant has data attached. Such enums are not yet supported." 
In this case, we would expect typescript bindings like:

```
type Foo = string | "Other"
```

### LINKED ISSUE #2: Serde tags + Enum compatability
Serde allows users to specify tags for the content and the tag of an enum.

For example,
```
#[derive(TS, Deserialize)]
#[serde(tag = "kind", content= "data")]
enum Biz {
    Foo,
    Bar
}
```

generates: `type Biz = "Foo" | "Bar"`

When really it should generate:
`type Biz = {kind: "Foo", content: null} | {kind: "Bar", content: null}`
The latter type is deserializable. The former is not.

jermdroui:
This pull requests Closes #1 and Closes #2.

This pull request is missing functionality for resolving enum dependencies.
I've also made some refactors to have greater reuse across enums/structs.


NyxCode:
fantastic work, thanks!

PR #4, opened 2021-04-05
More default derives

DusterTheFirst:
Currently, only `Option` and `Vec` derive `TS` by default. This PR adds derivations for `HashSet` and `BTreeSet` that match that of `Vec` since serde_json treats them the same. This PR also adds derivations for `HashMap` and `BTreeMap` both as dictionaries which is how serde_json would represent them.


NyxCode:
good work, thanks!

Issue #5, opened 2021-04-05
Ability to export to .d.ts files


DusterTheFirst:
When dealing with javascript code bases, it is sometimes useful to use typescript definition files even when you are not using typescript. Currently, the way the types are declared, they are `export`ed, meaning you must import them, like such
```js
/** @type {import("../types/generated/ingest.js").TelemetryPacket}  */
```
where as if they were `declared` ambiently, they would not need to be imported. This is common practice in typescript declaration files for non-modules. Below are examples of exports vs declarations.

This could be achieved through an attribute to the derive macro.

### Exports
![image](https://user-images.githubusercontent.com/14093962/113632756-bfcb3500-9639-11eb-88fe-ec8bc3caf6a6.png)

### Declaration
![image](https://user-images.githubusercontent.com/14093962/113632810-d2456e80-9639-11eb-91db-c41c6f9bf3a0.png)
(declarations are global, exports need to be `import`ed)

NyxCode:
I had not thought about using the definitions in vanilla js.
For typescript, I felt like [the consensous was that modules should be prefered over ambient defnintions](https://stackoverflow.com/questions/48690533/typescript-declare-vs-export-best-practices). 
Is the only difference the `declare` instead of `export`, or are there other differences besides that?

DusterTheFirst:
For my use case, the only difference is the word declare. 

It makes sense why the modules would be preferred, but when dealing with A library that just dumps its content globally or when dealing with JavaScript that doesn't have those import semantics built in (for older versions) it can be nice to have the option to use ambient declarations over the module system.

NyxCode:
Hey, I just added support for this in the `support-declare` branch.
```rust
/// By default, `export!` always uses `export type/interface`.
/// If you wish, you can also use ambient declarations (`declare type/interface`):
use ts_rs::{export, TS};

#[derive(TS)] struct Declared;
#[derive(TS)] struct Normal(Declared);

export! {
    (declare) Declared => "bindings/declared.d.ts",
    Normal => "bindings/normal.ts"
}

/// Since `Declared` is now an ambient declaration, `bindings/normal.ts` will not include an import
/// for `bindings/declared.d.ts`.
```

I'd appreciate if you could give it a try before I publish it to crates.io.
`ts-rs = { git = "https://github.com/Aleph-Alpha/ts-rs", branch = "support-declare" }`

DusterTheFirst:
[...]

PR #11, opened 2021-05-02
Support declare

NyxCode:
Add the ability to use `declare` instead of `export`.

```rust
export! {
  (declare) Role => "bindings/role.d.ts",
  User => "bindings/user.ts",
}
```


NyxCode:
This is a breaking change, since `TS::decl` doesn't include `export` aymore.  
Now, `export` or `declare` is added by the `export!` macro.

NyxCode:
resolves #5

PR #15, opened 2021-06-21
Implement internally tagged unions with subtypes

arendjr:
I have a follow-up PR :)

It now also allows for the use of subtypes inside internally tagged unions.

### DIFF
--- a/macros/src/types/mod.rs
+++ b/macros/src/types/mod.rs
@@ -102,9 +102,15 @@ fn format_variant(
                             #inline_flattened
                         )
                     },
-                    None => panic!(
-                        "Serde enums with tag discriminators should also have flattened fields"
-                    ),
+                    None => quote! {
+                        format!(
+                            "\n{}{{ {}: \"{}\" }} & {}",
+                            " ".repeat((indent + 1) * 4),
+                            #tag,
+                            #name,
+                            #inline_type
+                        )
+                    },
                 },
             },
             None => match &variant.fields {
--- a/macros/src/types/mod.rs
+++ b/macros/src/types/mod.rs
@@ -37,21 +37,32 @@ pub(crate) fn r#enum(s: &ItemEnum) -> Result<DerivedTS> {
     };
 
     let mut formatted_variants = vec![];
+    let mut dependencies = vec![];
     for variant in &s.variants {
-        format_variant(&mut formatted_variants, &enum_attr, &variant)?;
+        format_variant(
+            &mut formatted_variants,
+            &mut dependencies,
+            &enum_attr,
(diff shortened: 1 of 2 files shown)


NyxCode:
:D I was just looking at the same thing.
This is an interresting implementation..
I was thinking of generating something like
```ts
type MyEnum = { tag: "C", x: number, y: number }
```
from 
```rust
struct Point { x: u8, y: u8 }
enum MyEnum { C(Point) }
```
by properly implementing `TS::inline_flattened` for newtypes (currently, it just returns `None`, though I don't know why I did that :D)

arendjr:
Well, in many cases you might like to use the types of the variant independently, so I think it's nice to preserve the name of the inner types.

Still, if you compare it to manually written type definitions, it's probably a little awkward the discriminator is not part of the inner type. Ideally, I think we should generate this:

```ts
type A = { type: "A", foo: string };
type B = { type: "B", bar: number };
type MyEnum = A | B;
```

This would have a big advantage in consuming TypeScript code, because any object of type A or B can be directly passed wherever MyEnum is expected, without having to add the `type` anymore.

But I don't really see how we could do that, unless we want to introduce another custom annotation.

FWIW, we also use an OpenAPI-to-TS generator, which generates enums the same way this PR does now.

NyxCode:
I think we'd want the TS types to be generated like your PR does. While its maybe not the most elegant, it closely resembles the rust code and just seems "simpler" imho. 

I'm just a bit unsure about the API. Let me try to explain.
I noticed that I have not implemented inlining of newtype fields:
```rust
struct NewType(#[ts(inline)] Point);
```
This currently fails to compile, but should generate
```ts
type NewType = { x: number, y: number };
```
When I fix this, the `inline_flattened` field of the enum variant would not be `None` like it is right now.
So after this single line fix, this behaviour could be achieved by just using `#[ts(inline)]`:

```rust
struct MyEnum { A(#[ts(inline)] Point) } 
```
```ts
type MyEnum = {type: "A", x: number, y: number}
```
So I *think* this would be the cleaner fix, though I'm not sure..

NyxCode:
Although we could have both. Without #[ts(inline)], it generates your output, with #[ts(inline)], the the one.
Yeah, I think that's the way to go.
I guess I can merge the PR and then fix the newtype #[ts(inline)] implementation myself.

arendjr:
> Without #[ts(inline)], it generates your output, with #[ts(inline)], the [other] one.

Sounds good!

arendjr:
Added one more commit to correctly track dependencies for internally tagged unions.

NyxCode:
😘

Issue #14, opened 2021-06-21
Support for generics


arendjr:
```rs
#[derive(TS)]
struct A<T> {
    foo: String,
}
```

Results in the following error:

```
wrong number of type arguments: expected 1, found 0
expected 1 type argument
```

I suspect this error is triggered somewhere in the macro implementation, though I don't see where exactly. What would it take for generics to be supported? If I had a few pointers I might try to see if I could implement it myself.

PR #16, opened 2021-06-23
Generics support

arendjr:
This PR implements basic support for generics, resolving #14.

I am still very new to implementing these kinds of macros, and I'm well aware it's not the prettiest yet, so advice is welcome!

### DIFF
--- a/example/src/lib.rs
+++ b/example/src/lib.rs
@@ -37,11 +37,28 @@ enum Vehicle {
     Car { brand: String, color: String },
 }
 
+#[derive(Serialize, TS)]
+struct Point<T>
+where
+    T: TS,
+{
+    time: u64,
+    value: T,
+}
+
+#[derive(Serialize, TS)]
+struct Series {
+    points: Vec<Point<u64>>,
+}
+
 // this will export [Role] to `role.ts` and [User] to `user.ts` when running `cargo test`.
 // `export!` will also take care of including imports in typescript files.
 export! {
     Role => "role.ts",
     User => "user.ts",
+    // any type can be used here in place of the generic, as long as it impls TS:
+    Point<()> => "point.ts",
+    Series => "series.ts",
     Vehicle => "vehicle.ts",
     // this exports an ambient declaration (`declare interface`) instead of an `export interface`.
     (declare) Gender => "gender.d.ts",
--- a/macros/src/types/named.rs
+++ b/macros/src/types/named.rs
@@ -1,6 +1,9 @@
 use proc_macro2::TokenStream;
 use quote::quote;
-use syn::{Field, FieldsNamed, Result};
+use syn::{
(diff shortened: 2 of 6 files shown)


example/src/lib.rs
```diff
@@ -37,11 +37,28 @@ enum Vehicle {
     Car { brand: String, color: String },
 }
 
+#[derive(Serialize, TS)]
+struct Point<T>
+where
+    T: TS,
```
arendjr, line 43: I would have hoped we didn't need to specify this `where` clause manually, but I felt rewriting the `where` clause was a little bit too much to hope for in this first version...

example/src/lib.rs
```diff
@@ -37,11 +37,28 @@ enum Vehicle {
...
+{
+    time: u64,
+    value: T,
+}
+
+#[derive(Serialize, TS)]
+struct Series {
+    points: Vec<Point<u64>>,
```
arendjr, line 51: This one deserves a test on its own. While `Point<u64>` now passes the tests, embedding it inside a vector actually causes subtle breakage (the generic argument is no longer passed to TS). This might be useful for a follow-up PR, as I felt getting all edge-cases right in one go is currently out of my league.

example/src/lib.rs
```diff
@@ -37,11 +37,28 @@ enum Vehicle {
...
+}
+
 // this will export [Role] to `role.ts` and [User] to `user.ts` when running `cargo test`.
 // `export!` will also take care of including imports in typescript files.
 export! {
     Role => "role.ts",
     User => "user.ts",
+    // any type can be used here in place of the generic, as long as it impls TS:
```
arendjr, line 59: This feels a little bit like a work-around, but it's not our biggest problem :sweat_smile:

macros/src/types/named.rs
```diff
@@ -102,3 +179,18 @@ fn format_field(
...
+        || ident == "BTreeMap"
+        || ident == "BTreeSet"
+        || ident == "Cell"
+        || ident == "Cow"
+        || ident == "HashMap"
+        || ident == "HashSet"
+        || ident == "Option"
+        || ident == "Rc"
+        || ident == "RefCell"
+        || ident == "Vec"
```
arendjr, line 195: This looks really ugly, but I'm at a loss for how we can detect this at compile time and still have a graceful fallback...

example/src/lib.rs
arendjr: On second thought, if we did automatic rewriting, we wouldn't just need to rewrite it in the struct definition, but also for the `impl` blocks. That seems unreasonable, and I think it would also be very intuitive if you need to do manually for the `impl` blocks and *not* for the struct definition. So I'm starting to think it should stay like this.

arendjr:
Superseded by #18.

PR #17, opened 2021-06-28
Optional options

arendjr:
Currently, `ts-rs` translates `foo: Option<T>` to `foo: T | null` in TypeScript, but a common pattern is to use `#[serde(skip_serializing_if="Option::is_none")]` for types that are represented in TypeScript like this: `foo?: T`. It's a rather specific feature, but one that is very valuable for our usecase.

Note: This PR includes my other PRs, as I want to save myself the merge conflicts, so you may want to just look at the last commit. If necessary, I can split them out later.

### DIFF
--- a/example/src/lib.rs
+++ b/example/src/lib.rs
@@ -37,11 +37,28 @@ enum Vehicle {
     Car { brand: String, color: String },
 }
 
+#[derive(Serialize, TS)]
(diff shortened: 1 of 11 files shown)


arendjr:
Closing in favor of #19, which only contains the relevant commits.

PR #18, opened 2021-06-29
Generics support (part II)

### LINKED ISSUE #14: Support for generics

arendjr:
Fixes #14 and supersedes #16. I think I have managed to get support for generics to a level that it is more generically (heh) useful.

Generics can now be nested, and arrays and maps can be handled properly.


macros/src/types/named.rs
```diff
@@ -88,17 +117,118 @@ fn format_field(
...
+                .iter()
+                .map(|ty| format_type(ty, generics))
+                .collect::<Vec<_>>();
+            let args = quote!(vec![#(#args),*]);
+            quote!(<#ty as ts_rs::TS>::name_with_type_args(#args))
+        }
+    }
+}
+
+fn has_specialized_impl(ident: &Ident) -> bool {
```
arendjr, line 226: This one is not as bad as during the first attempt (because arrays and maps are now handled more generically), but I'm still open to suggestions on how to improve this.

arendjr:
Oops. I tried to fix the merge conflict, and merged in my own `main` instead of your upstream one, so now this PR contains *everything*. I'll close it again and reopen to limit it to Generics again :)

PR #19, opened 2021-06-29
Optional Options (part II)

arendjr:
Currently, `ts-rs` translates `foo: Option<T>` to `foo: T | null` in TypeScript, but a common pattern is to use `#[serde(skip_serializing_if="Option::is_none")]` for types that are represented in TypeScript like this: `foo?: T`. It's a rather specific feature, but one that is very valuable for our usecase.

(I have recreated this PR so it doesn't include commits from other PRs)


NyxCode:
great work, thanks!

PR #20, opened 2021-06-30
Generate index

arendjr:
This implements the ability for the `export!` macro to generate an index file that simply re-exports all the generated type definitions.

Example:

```rust
use ts_rs::{export, index};

export ! {
    A => "bindings/a.ts",
    index => "bindings/index.ts" // export type { A } from "./a.ts";
}
```

PS.: This might be my last PR for a little while, as it appears our use case has now been successfully implemented :)

### DIFF
--- a/ts-rs/src/export.rs
+++ b/ts-rs/src/export.rs
@@ -5,6 +5,20 @@ use std::path::{Component, Path, PathBuf};
 use crate::TS;
 use std::fmt::Write;
 
+/// Placeholder type that can be used to instruct `export!` to generate an index file that
+/// re-exports all generated types.
+///
+/// ```rust
+/// use ts_rs::{export, index};
+///
+/// export ! {
+///     A => "bindings/a.ts",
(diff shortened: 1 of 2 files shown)


NyxCode:
:D Thanks a lot for all the PRs, It'll take me a few days to look at them all.

NyxCode:
While I like the idea of generating an index file, I'm not sure using an `index` type for this is the best approach. I'll take a look at how to implement this within the `export!` macro.

arendjr:
Sure! I was also not entirely sure what the best approach would be, but I settled for this because I liked how it fits with the existing model and allows you to the configure the path the same way. But I did feel it required a bit of work-around in the implementation, so I can understand if you prefer something else instead.

NyxCode:
in v6.0.0, the `export!` macro was replaced by `#[ts(export)]` to support generating imports without having to list all types in a single invocation of `export!`. 
With the new setup, generating an index file would have to be done seperately - maybe by hand or by an external tool. There is sadly no good way to keep state between multiple invocations of a proc macro right now.

PR #21, opened 2021-06-30
Sort imports

arendjr:
Replace the usage of `HashMap` and `HashSet` with `BTreeMap` and `BTreeSet`. This way, imports are automatically sorted, which has the important benefit of making them stable across `cargo test` invocations.


NyxCode:
Good idea, thanks!

PR #24, opened 2021-07-19
add #[ts(type=..)] on enums for `enum` generation

### LINKED ISSUE #23: Export `enum` instead of `type` for simple C-like enums
When Rust's enums could be represented with TypeScript's enums I think they should. So for example an enum like:

```rs
#[derive(serde::Serialize, ts_rs::TS)
enum Color {
	#[serde(rename="R")
	Red,
	Blue,
	Green
}
```

should end up as

```ts
export enum Color {
	Red="R",
	Blue,
	Green
}
```

Maybe even `const enum`? Should probably be optional with an attribute on the enum.

Instead of currently it being: `export type Color = "R" | "Blue" | "Green"`;

JakubKoralewski:
Previously having even a simple enum would generate a `type`. The default behavior does not change, but if specifed to be `enum` or `const enum` the corresponding will be generated.

Closes #23

### DIFF
--- a/macros/src/attr/enum.rs
+++ b/macros/src/attr/enum.rs
@@ -10,6 +10,7 @@ pub struct EnumAttr {
     pub tag: Option<String>,
     pub untag: bool,
     pub content: Option<String>,
+    pub r#type: Option<String>
 }
 
 #[cfg(feature = "serde-compat")]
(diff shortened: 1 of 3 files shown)


jy1263:
i'd love for something like this to be added! mayb could also implement serde_repr integration?

lorenzolewis:
Adding on that I'd love to see this added in. I just ran into a scenario where having this adjusted would be perfect for my use case!

GhaithAlHallak8:
Any updates?

escritorio-gustavo:
Closed as discussed in #118

Issue #22, opened 2021-07-09
Improve support for enum variants


arendjr:
I noticed a few issues with enum variants. For instance:

```rs
#[derive(Serialize, TS)]
#[serde(tag = "kind", content = "data")]
enum ComplexEnum {
    A,
    B { foo: String, bar: f64 },
    W(SimpleEnum),
    F { nested: SimpleEnum },
    T(i32, SimpleEnum),
    V(Vec<Series>),
}
```

generates the following type:

```ts
// issue 1: there is no import for the `SimpleEnum` type

export type ComplexEnum =
  | { kind: "A"; data: null } // issue 2: IMO it would be better to omit the `data` property
  | {
    kind: "B";
    data: {
      foo: string;
      bar: number;
    };
  }
  | { kind: "W"; data: SimpleEnum }
  | {
    kind: "F";
    data: {
      nested: SimpleEnum;
    };
  }
  | { kind: "T"; data: [number, SimpleEnum] }
  | { kind: "V"; data: Array }; // issue 3: there is no type argument passed to `Array`
```

I have fixed this issues in my own fork with the following commit: https://github.com/arendjr/ts-rs/commit/e7a87b1b49bfa0263460a63ab995fc133a4ba362

Unfortunately, it would be a bit unwieldy to create a PR for this as this is built on top of both #15 and #18. This is just to let you know these issues are addressed and I can create a PR for it when the other PRs are merged.

arendjr:
As #15 has now been merged, I have included the fixes for these issues in #25, since they partially affected generics support anyway.

arendjr:
Fixed as part of #25.

PR #25, opened 2021-08-04
Generics support (part III)

### LINKED ISSUE #14: Support for generics

arendjr:
This is the latest PR for applying generics support. It fixes #14 and #22 and supersedes #18.

Generics are now supported, which includes generic structs such as `MyStruct<T>` and generic arguments to existing containers such as `Vec<MyItem>`.

It also contains a few minor fixes as described in #22.

### DIFF
--- a/example/Cargo.toml
+++ b/example/Cargo.toml
@@ -6,4 +6,4 @@ edition = "2018"
 
 [dependencies]
 ts-rs = { path = "../ts-rs" }
-serde = { version = "1", features = ["derive"] }
+serde = { version = "1", features = ["derive", "rc"] }
--- a/example/src/lib.rs
+++ b/example/src/lib.rs
@@ -1,6 +1,8 @@
 #![allow(dead_code)]
 
 use serde::Serialize;
+use std::collections::BTreeSet;
+use std::rc::Rc;
 use ts_rs::{export, TS};
 
 #[derive(Serialize, TS)]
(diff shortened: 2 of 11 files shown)


marjakm:
I added a bit on top of your branch @arendjr 

This commit https://github.com/bislydev/ts-rs/commit/d7afa2f1881199f27a5c8b97254d0b4325ab94f4

Makes this work:

```rust
#[derive(TS)]
struct Generic2<T> {
    value: T,
    values: Vec<T>,
}
```

Without it, you had to do this (but in rust you dont ususally add type bounds to data):

```rust
#[derive(TS)]
struct Generic<T>
where
    T: TS,
{
    value: T,
    values: Vec<T>,
}
```

That commit can be improved on, it currently only adds bounds when there are none, but it could also insert bounds if there already are some. This branch could be implemented https://github.com/bislydev/ts-rs/commit/d7afa2f1881199f27a5c8b97254d0b4325ab94f4#diff-4feb5862b96e70d6a225028210762f9a2d19ea5e35860fe78d211fb607c82b26R82

I'm a bit busy at the moment, but could help with implementing it at some later time

NyxCode:
Hm, how should we procede with this? I'd happily merge this as-is, or should we wait for @marjakm additions?

marjakm:
I created a pull request to the @arendjr branch this pull request is based on. https://github.com/arendjr/ts-rs/pull/1
You could also just merge this as-is and I could make the pull request against this repo.

arendjr:
Nice job! I’ve merged. Just need to resolve some conflicts, I see.

arendjr:
All merged, and tests passed. Should be good to go!

NyxCode:
Finally! 😉 Thanks for your awesome work, and sorry that it took so long.

NyxCode:
0.4 is out!

NyxCode:
@arendjr I found a regresion when using a generic type within `Option`:
```rust
    #[derive(TS)]
    struct List {
        data: Option<Vec<u32>>
    }
```
results in 
```ts
interface List {
  data: Array | null
}
```

NyxCode:
I have not gone super deep into the implementation, but I would guess that Option should now be implemented like this:
```rust
impl<T: TS> TS for Option<T> {
    fn name() -> String {
        unreachable!();
    }

    fn name_with_type_args(args: Vec<String>) -> String {
        assert_eq!(args.len(), 1);
        format!("{} | null", args[0])
    }
}

arendjr:
You're right, that seems indeed the proper fix. Thanks!

PR #31, opened 2021-09-28
Feat/UUID

Roms1383:
This is a minor addition for [uuid::Uuid](https://crates.io/crates/uuid) which just map it to a `string`.
It's gated behind a `uuid-impl` feature, similar to `chrono-impl`.

I also added it to the example :)


NyxCode:
thanks!

Issue #71, opened 2021-12-29
Wrong import paths in generated TS files for projects using Cargo Workspaces


FruitieX:
This problem occurs in a project where we have put the project's Rust types into their own crate, so that these types can be shared and used both from e.g. backend and frontend crates.

Minimal repro: https://github.com/FruitieX/ts-rs-repro

Here we have two crates `app` and `types`. `types` contains a SharedType struct that's exported via ts-rs. `app` uses this type as part of it's internal AppState type, which I also want to export to TS using ts-rs.

The typings can be generated by running `cargo test` in the root of the repo. I've committed the generated bindings for convenience.

The TS typings get exported into `app/bindings/AppState.ts` and `types/bindings/SharedType.ts` respectively. However, `app/bindings/AppState.ts` tries to import from the relative path `./SharedType`, which doesn't work since the actual working relative path would be `../../types/bindings/SharedType`.

I'm currently working around this by manually moving the generated bindings into a shared `bindings` dir in the root of our project repo.

NyxCode:
Crrently, ts-rs exports to `$CARGO_MANIFEST_DIR/{configured path}`, which breaks here.

NyxCode:
When you export a few types in the parent directory using `#[ts(export_to = "../bindings")]`, does import resolution for those types work like one would expect?

FruitieX:
That would be a good solution for me, but unfortunately it seems to generate a single file called `bindings` in the root of the repo. And this file will only contain one of the generated types, presumably the one that was processed last.

If I instead specify the full paths like
`#[ts(export, export_to = "../bindings/AppState.ts")]` and 
`#[ts(export, export_to = "../bindings/SharedType.ts")]` respectively in the repro example, then import resolution will work correctly as they are placed in the same output directory but in separate files. In the repro example doing this is fine, but if you have loads of types to export it might get tedious.

Would it be desirable that the `export_to` attribute looks at the path to figure out whether it containsh a filename extension or not, and if no filename extension is found it writes the output to the given directory but in files named according to the type? I could try creating a pull request to do that if wanted.

NyxCode:
> If I instead specify the full paths like

Ah yeah that's what I meant.  
Yeah, that'd be great. I think we shouldn't look at the extension, but if the path contains a trailing `/`. `/some/path` should be treated as a file, but `/some/path/` as a directory.  

If you need any guidance regarding the PR, let me know!

NyxCode:
[...]

PR #73, opened 2022-01-03
export_to supports output to directories

FruitieX:
Related to discussion in #71.

Adds support for output to directories when the export_to path contains a trailing `/` character.


NyxCode:
thanks!

PR #77, opened 2022-01-19
Add support for serde(tag=...) on structs

HalfVoxel:
Serde supports this (see https://github.com/serde-rs/serde/pull/1448).  Though serde's documentation is missing it at the moment.
It is very useful for bringing things to typescript.


NyxCode:
nice work!

PR #79, opened 2022-01-20
Disable Inflector default features

asv7c2:
Removes regex and lazy_static from dependency tree.
---
Adds support for Result enum.


NyxCode:
I don't think translating `Result<T, E>` to `E` is appropriate here.  
Maybe we could split this into 2 PRs, one for the default features and one for `Result`?

asv7c2:
@NyxCode 
Updated. Sorry.

PR #88, opened 2022-03-14
add support for indexmap's `IndexMap` and `IndexSet` types under a `i…

rhsmits91:
…ndexmap-impl` feature flag


NyxCode:
cool, thx

Issue #89, opened 2022-03-21
Ability to set output path on TS::export


teenjuna:
In my application, I have a separate CLI command to generate types. I would like to be able to set output file inside the implementation of this command. However, right now the only way to specify output path is by using `#[ts(export_to = "...")]` attribute. 

My suggestion: introduce new method `export_to(path: &Path)` to the `TS` trait. This method does the same thing as the `TS::export()`, but exports to the specified path. Shouldn't even be a breaking change!

PR #90, opened 2022-03-21
Add `TS::export_to` method

teenjuna:
Solves #89


NyxCode:
Thanks for the PR.
Something,  I assume you use IntelliJ/CLion, removed the necesarry trailing whitespaces in the doc comments.  
Could you undo that change? If you do use IntelliJ/CLion, you can disable this behaviour in the settings.

NyxCode:
merged this manually instead

teenjuna:
I'm sorry, I totally forgot about this... I didn't use IntelliJ/CLion though, it was just Neovim with standard `rustfmt` on save.

PR #97, opened 2022-05-01
Implement ordered-float support

theduke:
Adds support for the ordered-float crate with a new "ordered-float-impl"
feature.


NyxCode:
Could you also add this in the doc comment in lib.rs and re-generate the README.md from it (see the contributing file)?

PR #98, opened 2022-05-01
Add export_to_string functionality

theduke:
Adds  a new method TS::export_to_string() that allows getting the
generated defintion as a string.


NyxCode:
👍 thanks

PR #99, opened 2022-05-01
Wrap invalid field names in quotes

theduke:
If a struct field name would be an invalid Typescript ident,
the field name is wrapped in quotes.


NyxCode:
I like this! Maybe it should even be the default behaviour for every identifier - what do you think?

PR #101, opened 2022-05-01
Add feature to toggle export test codegen

theduke:
Adds a new feature `text-export` that toggles the code generation of export
tests.

Feature is on by default.

### DIFF
--- a/macros/Cargo.toml
+++ b/macros/Cargo.toml
@@ -10,6 +10,7 @@ repository = "https://github.com/Aleph-Alpha/ts-rs"
 
 [features]
 serde-compat = ["termcolor"]
+test-export = []
 
 [lib]
 proc-macro = true
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -58,7 +58,13 @@ impl DerivedTS {
         };
 
         let export = match self.export {
-            true => Some(self.generate_export_test(&rust_ty, &generics)),
+            true => {
+                if cfg!(feature = "test-export") {
+                    Some(self.generate_export_test(&rust_ty, &generics))
+                } else {
+                    None
+                }
+            }
             false => None,
         };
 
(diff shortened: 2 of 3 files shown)


macros/src/lib.rs
```diff
@@ -58,7 +58,13 @@ impl DerivedTS {
         };
 
         let export = match self.export {
-            true => Some(self.generate_export_test(&rust_ty, &generics)),
+            true => {
+                if cfg!(feature = "test-export") {
```
NyxCode, line 62: The following would be more elegant here:
```rust
let export = match self.export {
    #[cfg(feature = "test-export")]
    true => { .. }
    _ => None
```
This would also work:
```rust
```rust
let export = match self.export {
    true if cfg!(feature = "test-export") => { .. }
    _ => None
```

```

File: macros/Cargo.toml [hunk 2]
```diff
@@ -10,6 +10,7 @@ repository = "https://github.com/Aleph-Alpha/ts-rs"
 
 [features]
 serde-compat = ["termcolor"]
+test-export = []
```
NyxCode, line 13: I think i'd prefer a more descriptive name, maybe something like "export-on-test"

ts-rs/Cargo.toml
```diff
@@ -12,14 +12,18 @@ categories = ["development-tools::ffi", "development-tools::build-utils", "wasm"
 readme = "../README.md"
 
 [features]
+# If enabled, each derived type will be exported in an automaticaly generated test.
```
NyxCode, line 15: This is not entirely correct, since tests are only generated when `#[ts(export)]` is present. I think we should highlight in the comment that this is meant for toggling this on and off

NyxCode:
Could you also add this to the list of cargo features in the doc comment in lib.rs, and re-generate the README.md?

PR #102, opened 2022-05-05
Implementing typescript doc_string

Mattchine:
Example in test file.

Fix my [ts-to-zod issue](https://github.com/fabien0102/ts-to-zod/issues/80).

Ps. I don't think passing comment from `rust` to `ts` is necessary so I decided to implement it as docstring


NyxCode:
thanks for the PR! Could you take a look at the merge conflict? Once it's resolved, I'd love to merge this.

Mattchine:
I think this solve the conflict.

NyxCode:
@Mattchine Sorry for the late reply. I currently don't have much time for the project, though I hope that will change soon.  

> I have an idea of taking multi-line docstring as
`#[ts(doc_string=["@params a", "second line"])]`
But I'm not sure how to implement this. Any idea?

I think it should be possible to parse vanilla doc-comments (`/// ..`) in the proc macro. [As far as I know, `///` is desugared into `#[doc = ".."]`](https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html). I'll look into this as soon as I've got some spare time. 

I feel like it would make sense to use a similar syntax for `#[ts(doc_string)]`. I also think it might make sense to directly parse `#[doc]` attributes and convert them to TS doc strings, but I'm not completely sure about that yet.

timephy:
This would be awesome!
What's the state of this PR? I'd love to be able to work with this soon..

If you feel uncertain about whether to translate all docs from RS to TS - how about:

- `ts(doc)` translates docs from RS to TS.
- `ts(doc_string="...")` puts the given string into TS.
- Otherwise no docs in TS

escritorio-gustavo:
This has been inactive since 2022 and #187 seems to reimplement it, so I'll be closing this PR

PR #103, opened 2022-05-13
Support PhantomData wrapper type

FruitieX:
This PR adds support for std::marker::PhantomData<T>.

Useful for situations where you might have some type information associated with a struct, but that's not actually part of the struct's data.

There's one problem with this currently, and that is that the TypeScript typings are generated as if the PhantomData contents could be read directly. Reading from the field should be blocked somehow, but I'm not really sure how to accomplish that in TypeScript while also conveniently being able to access the type information.


NyxCode:
I merged this as-is. Right now, one would probably want to `#[ts(skip)]` these fields though.  
In the future, it'd be nice if these fields would be ignored automatically.

PR #104, opened 2022-05-13
Fix compilation errors when building with `format` feature

FruitieX:
The `path` variable is not available in this function (anymore?), and as such compilation will fail. Fix this by feeding a dummy path to `format_text()`. I'm not sure if there are any wider implications to this, but it seems to work in my testing.


NyxCode:
Hey! Sorry, this seems to have slipped through while merging a PR. Will fix it asap.

NyxCode:
fixed on master, release coming soon(tm)

PR #114, opened 2022-08-18
Implement `char` as a primitive

SplittyDev:
Not much to see here, just added `char` as a primitive binding to `string`.


SplittyDev:
Tbh I'm not 100% sure whether chars should be stored as a `string` or `number`, but so far I've been doing it with `string` and it worked fine. Not 100% correct on the TypeScript side since you _could_ specify a multi-character string, but in that case `number` also wouldn't be semantically correct, so I think for now `string` is the most accurate representation.

NyxCode:
I feel like serializing it as `string` is the least unintuitive, and, more importantly, how serde_json does it.  
Thanks for the PR!

PR #115, opened 2022-09-13
Support kebab-case rename

PhotonQuantum:
According to serde document, the possible values for `rename_all` are "lowercase", "UPPERCASE", "PascalCase", "camelCase", "snake_case", "SCREAMING_SNAKE_CASE", "kebab-case", "SCREAMING-KEBAB-CASE".

This crate now supports the first six variants, but `kebab-case` is not one of them. So this PR adds it to the list.

Unfortunately, `SCREAMING-KEBAB-CASE` is not supported by the Inflector crate. So I leave it as is.

### DIFF
--- a/macros/src/attr/mod.rs
+++ b/macros/src/attr/mod.rs
@@ -20,6 +20,7 @@ pub enum Inflection {
     Snake,
     Pascal,
     ScreamingSnake,
+    Kebab,
 }
 
 impl Inflection {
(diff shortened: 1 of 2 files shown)


NyxCode:
Awesome, thanks!

NyxCode:
Can you remove the unused import? Otherwise, I can merge it as-is and fix that later.

PhotonQuantum:
Sorry! My IDE auto-completed the redundant import. Fixed.

PR #117, opened 2022-09-14
Hackily support non-'static lifetimes

petersn:
This is a hacky attempt to make ts-rs support non-`'static` lifetimes by simply dropping the quantification over lifetimes in our `impl`s of `TS`, and remapping every lifetime referenced inside of the `impl` to be `'static`. This PR seems to pass the tests, but I'll admit that I just hacked this together in a couple hours without carefully looking over ts-rs, so I have no idea if it breaks everything. Almost certainly this deserves more careful scrutiny by somebody who didn't just come to ts-rs's source for the first time today.

Here is an example struct that didn't work before, but works now:
```
#[derive(ts_rs::TS)]
#[ts(export)]
struct Foo<'a> {
    field: &'a str,
}
```

(This correctly exports to `export interface Foo<> { field: string, }`.)

### DIFF
--- a/macros/Cargo.toml
+++ b/macros/Cargo.toml
@@ -17,6 +17,6 @@ proc-macro = true
 [dependencies]
 proc-macro2 = "1"
 quote = "1"
-syn = { version = "1", features = ["full", "extra-traits"] }
+syn = { version = "1", features = ["full", "extra-traits", "visit-mut"] }
 Inflector = { version = "0.11", default-features = false }
 termcolor = { version = "1", optional = true }
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -112,30 +112,26 @@ impl DerivedTS {
 fn generate_impl(ty: &Ident, generics: &Generics) -> TokenStream {
     use GenericParam::*;
 
-    let bounds = generics.params.iter().map(|param| match param {
+    let bounds = generics.params.iter().filter_map(|param| match param {
         Type(TypeParam {
             ident,
             colon_token,
             bounds,
             ..
-        }) => quote!(#ident #colon_token #bounds),
-        Lifetime(LifetimeDef {
-            lifetime,
-            colon_token,
-            bounds,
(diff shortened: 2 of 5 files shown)


NyxCode:
Thanks for your contribution, I really appreciate it. I will take a closer look in the coming days.

petersn:
I have four other changes on my branch:
1. I also made lifetimes work in anonymous fields of structs -- I should probably cherry-pick that commit into this PR.
2. I made slices turn into `Array`s, because currently a field like `&'a [i32]` doesn't work.
3. Currently `ts-rs` doesn't handle `rename_all` attributes applied to enum variants, resulting in the TypeScript it outputs being incompatible with what Serde outputs. I have a fix for this.
4. Currently `skip_serializing_if` only works if the argument is `Option::is_none`, and the TypeScript unwraps the `Option`, to avoid also having a `| null`. I made it emit a ? TypeScript field if it's something else that's unknown. I'm not totally clear on what the right behavior is here.
I should probably make another PR, or make several, I assume?

thomasmost:
@NyxCode Just wanted to +1 this, if you have bandwidth!

petersn:
@thomasmost In case you're trying to use this right now, you can also include in your Cargo.toml via:

```ts-rs = { git = "https://github.com/petersn/ts-rs", branch = "optional-changes" }```

This should give you all of the above mentioned features (non-static lifetimes, slices becoming arrays, `rename_all` on enum variants, and `skip_serializing_if` behavior that's more consistent with serde).

escritorio-gustavo:
I believe #181 (and possibly other PRs) have fixed this, as the example provided no longer requires `'a: 'static`

Slices now work as the PRs #183 and #199 were directed at fixing them

`rename_all` has worked on enum variants for some time, though I don't know when that was fixed (see #135)

I haven't checked the `skip_serializing_if` behavior, but I do believe `serde` converts `None` to `null`, so `| null` is likely correct behavior

PR #118, opened 2022-09-16
feat: reimplement #[type = ..] on enums for `enum` generation

pintariching:
This is a reimplementation of #24. I tried merging that PR into main but I had some merge issues because the PR is old. 
All the tests pass and you can now add `#[ts(type = "enum"]` or `#[ts(type = "const enum"]` to generate an enum like 
```rust
#[ts(type = "const enum")]
enum Dog {
  Woof,
  Bark,
  Howl
}
```
generates a typescript enum like 
```typescript
const enum Dog = { Woof, Bark, Howl };
```
Credits to @JakubKoralewski for implementing this.

### DIFF
--- a/macros/src/attr/enum.rs
+++ b/macros/src/attr/enum.rs
@@ -11,9 +11,10 @@ pub struct EnumAttr {
     pub rename: Option<String>,
     pub export_to: Option<String>,
     pub export: bool,
-    tag: Option<String>,
-    untagged: bool,
-    content: Option<String>,
+    pub r#type: Option<String>,
+    pub tag: Option<String>,
+    pub untagged: bool,
+    pub content: Option<String>,
 }
 
 #[cfg(feature = "serde-compat")]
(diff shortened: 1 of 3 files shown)


NyxCode:
Thanks, really appreciate it! The tests currently seem to fail, do you know why?

pintariching:
The problem seems to be running `cargo test --no-default-features`. I'll check it out.

pintariching:
Ok I think I fixed it. The tests were using attribute macros from `serde`like `serde(rename_all = ...)` instead of `ts(rename_all = ...)` without having having the right directives enabled. I changed them to the latter now it should pass.

lorenzolewis:
@NyxCode is there any possibility for this one to be merged in when you have capacity?

escritorio-gustavo:
@NyxCode  This PR hasn't been active since 2022. Considering that and the fact that TS enums are generally considered to be really bad, should this PR be closed?

NyxCode:
@escritorio-gustavo Agreed, especially with your current work on enums in mind, which might require some deep refactors.  
If @pintariching is still interested in this, it'd probably make sense to re-write this based on the current master once enum flattening lands.

escritorio-gustavo:
Great, given that I'll also close #24, since this is a reimplementation of that

PR #120, opened 2022-09-30
TypeGroup support in `extract_type_args`

blackbeam:
The issue this PR fixes:

It turns out that field type is wrapped into a `Type::Group` in some cases (e.g when struct definition is generated via macro) and `extract_type_args` is unable to properly extract types.


NyxCode:
thanks!

Issue #119, opened 2022-09-20
Support #[serde(rename)] for enum variants


kimpers:
Thanks for an awesome library. However, the lack of `#[serde(rename)]` support is blocking me from using it in production.

**Example**
```rust
#[derive(Serialize, Deserialize, TS)]
#[serde(tag = "kind")]
#[ts(export)]
pub enum Shapes {
    #[serde(rename = "SQUARE_THING")]
    SquareThing {
        name: String,
        // ...
    },
}
```

**Warning**
```
warning: failed to parse serde attribute
  |
  | #[serde(rename = "SQUARE_THING")]
  |
  = note: ts-rs failed to parse this attribute. It will be ignored.
```
**Actual result**
The types have en incorrect TS enum
```typescript
export type SquareThing =
  | {
      kind: "squareThing";
      name: string
    };
```

**Expected result**
```typescript
export type SquareTing =
  | {
      kind: "SQUARE_THING";
      name: string
    };
```

phoneticsushi:
Independently encountered a similar issue.  This seems highly related to (if not a duplicate of) #72 - might be worth subscribing to that one.

phoneticsushi:
Hi again - I was trying to resolve the issue with `#[serde(rename_all)]`, which doesn't work, but it looks like `#[serde(rename)]` works fine:

```rust
#![allow(dead_code)]

use serde::{Serialize, Deserialize};
use ts_rs::TS;

#[derive(Serialize, Deserialize, TS)]
#[serde(tag = "kind")]
#[ts(export)]
pub enum Shapes {
    #[serde(rename = "SQUARE_THING")]
    SquareThing {
        name: String,
        // ...
    },
}

#[test]
fn test_enum_variant_with_tag() {
    assert_eq!(
        Shapes::inline(),
        "{ kind: \"SQUARE_THING\", name: string, }"
    );
}
```
Have you tried this again recently?  The above works on commit `b79cb2a274df40da19686fee65f1142f5da4e272`

NyxCode:
A new release is long overdue :D 
But yes, it works on master.

Issue #72, opened 2022-01-01
Support #[serde(rename_all)] for enum variants


Simon-Laux:
Example:
 ```rs
use ts_rs::TS;
use serde::Serialize;

#[derive(Serialize, TS)]
#[serde(tag = "type")]
pub enum Thing {
    #[serde(rename_all = "camelCase")]
    A {
        display_name: Option<String>,
    },
    B (u8),
}
```
resulting warning:
```
warning: failed to parse serde attribute
  | 
  | #[serde(rename_all = "camelCase")]
  | 
  = note: ts-rs failed to parse this attribute. It will be ignored.
```

so the resulting variant does not have it's field-names renamed in camel casing.

see https://serde.rs/attributes.html

phoneticsushi:
I encountered this independently.  Took a crack at it, but couldn't quite do it.

The immediate issue is in the `Parse` function called by the `TryFrom` implementation of the `SerdeEnumAttr` class:
https://github.com/Aleph-Alpha/ts-rs/blob/b79cb2a274df40da19686fee65f1142f5da4e272/macros/src/utils.rs#L99

...where both `Parse` and `TryFrom` are implemented by this macro:
https://github.com/Aleph-Alpha/ts-rs/blob/b79cb2a274df40da19686fee65f1142f5da4e272/macros/src/utils.rs#L15

...which is invoked here:
https://github.com/Aleph-Alpha/ts-rs/blob/b79cb2a274df40da19686fee65f1142f5da4e272/macros/src/attr/enum.rs#L83-L92

Expanding that invocation gives this - note the arms in the `match` block:
```rust
impl syn::parse::Parse for SerdeEnumAttr {
    fn parse(input: syn::parse::ParseStream) -> syn::Result<Self> {
        let mut out = SerdeEnumAttr::default();
        loop {
            let key: Ident = input.call(syn::ext::IdentExt::parse_any)?;
            match &*key.to_string() {
                "rename" => (out.0.rename = Some(parse_assign_str(input)?)),
                "rename_all" => (out.0.rename_all = Some(parse_assign_inflection(input)?)),
                "tag" => (out.0.tag = Some(parse_assign_str(input)?)),
                "content" => (out.0.content = Some(parse_assign_str(input)?)),
                "untagged" => (out.0.untagged = true),
                #[allow(unreachable_patterns)]
                _ => {
                    return Err(syn::Error::new((input.span()), {
[...]

PR #122, opened 2022-10-12
Support serde's "rename_all" on enum variants

phoneticsushi:
Adds tests and implementation resolving #72 and #119, though the latter was already working AFAICT.

### DIFF
--- a/macros/src/attr/mod.rs
+++ b/macros/src/attr/mod.rs
@@ -7,10 +7,12 @@ use syn::{
     parse::{Parse, ParseStream},
     Error, Lit, Result, Token,
 };
+pub use variant::*;
 
 mod r#enum;
 mod field;
 mod r#struct;
+mod variant;
 
 #[derive(Copy, Clone, Debug)]
 pub enum Inflection {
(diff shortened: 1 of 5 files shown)


phoneticsushi:
This PR is failing tests, but that same commit ([a6825f3](https://github.com/Aleph-Alpha/ts-rs/pull/122/commits/a6825f3fcc442243406e0dae257d8bb04ac1d35a)) passes `cargo clean; cargo test` on my machine and correctly generates typescript files in camelCase when included as a dependency of a separate project.  Any suggestions are appreciated

NyxCode:
Hey!
The issue seems to be that the test is run once with `cargo t` and once with `cargo t --no-default-features`.  
I'd recommend you 
1. not only test the `#[serde(..)]` annotations, but also the equivalend `#[ts(..)]` annotations.
2. Put tests which rely on the `serde-compat` feature behind a `#[cfg(feature = "serde-compat")]` gate

NyxCode:
For example, this would do the trick:
```rust
    #[cfg_attr(feature = "serde-compat", serde(rename_all = "camelCase"))]
    #[cfg_attr(not(feature = "serde-compat"), ts(rename_all = "camelCase"))]
    ```

NyxCode:
let me just merge this as-is and apply the minor fix to the tests. Thanks, really appreciate it!

PR #125, opened 2022-10-20
support std::net types

### LINKED ISSUE #110: Support IpAddr's
I am currently using `Ipv4Addr` in a project with `ts_rs` and found that is is not supported. It would be really nice if `ts_rs::TS` was implemented for:
 - [`std::net::Ipv4Addr`](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html)
 - [`std::net::Ipv6Addr`](https://doc.rust-lang.org/std/net/struct.Ipv6Addr.html)
 - [`std::net::IpAddr`](https://doc.rust-lang.org/std/net/enum.IpAddr.html)
 - [`std::net::SocketAddrV4`](https://doc.rust-lang.org/std/net/struct.SocketAddrV4.html)
 - [`std::net::SocketAddrV6`](https://doc.rust-lang.org/std/net/struct.SocketAddrV6.html)
 - [`std::net::SocketAddr`](https://doc.rust-lang.org/std/net/enum.SocketAddr.html)

I am currently unable to do this as a PR due to time constraints but if some time frees up I will drop a message and give it a go.

NyxCode:
resolves #110

### DIFF
--- a/ts-rs/src/lib.rs
+++ b/ts-rs/src/lib.rs
@@ -136,7 +136,7 @@ use std::{
     any::TypeId,
     collections::{BTreeMap, BTreeSet, HashMap, HashSet},
     ops::{Range, RangeInclusive},
-    path::{Path, PathBuf},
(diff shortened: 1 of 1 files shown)


NyxCode:
serde-json maps them to a string, so we'll do the same here

PR #127, opened 2022-10-24
Use type overrides for newtype enum variants in internal and adjacent representations

### LINKED ISSUE #126: Type overrides ignored for newtype enum variants in internal and adjacent representations
Hi,
Thanks for your work on ts-rs.

Currently the `#[ts(type = "something")]` overrides are not used for newtype variants for enums that use internal or adjacent representations:

```rs
#[derive(Serialize)]
struct S;

// compiles
#[derive(TS, Serialize)]
enum External {
    Newtype(#[ts(type = "unknown")] S),
}

// error[E0277]: the trait bound `S: TS` is not satisfied
#[derive(TS, Serialize)]
#[serde(tag = "tag")]
enum Internal {
    Newtype(#[ts(type = "unknown")] S),
}

// error[E0277]: the trait bound `S: TS` is not satisfied
#[derive(TS, Serialize)]
#[serde(tag = "tag", content = "content")]
enum Adjacent {
    Newtype(#[ts(type = "unknown")] S),
}
```
The code that causes this behaviour (for adjacent enums) is at https://github.com/Aleph-Alpha/ts-rs/blob/d59ca01878ce03f7b62ff0be56b350808fd257ed/macros/src/types/enum.rs#L97-L100

The simplest fix would probably be to parse the field attributes and just output the override if there is one.

Heliozoa:
Closes #126 

I wrote this quickly so there may be some side-effects I haven't considered, or maybe ignoring all of the other attributes is still problematic.


Heliozoa:
@NyxCode Tagging you just in case, since the other PRs got merged I've also updated this one as it had a merge conflict and a suboptimal test.

NyxCode:
thanks!

PR #128, opened 2022-10-24
Alleviate 'static/Sized restrictions

Heliozoa:
Hi,
Thanks for your work on ts-rs.

Here's my take on alleviating some of the 'static and Sized related restrictions for TS.

- 'static: This PR would allow non-'static types to derive TS and use decl and such, only gating the dependencies/export* and other functions that actually need it behind the 'static requirement.
- Sized: Allows the T inside Box, Rc, Arc and Cow to be unsized, such as in `Cow<'a str>`.

This makes some of the generic bounds more complicated and may potentially cause some confusion as to why exporting a non-'static type that implements TS doesn't work, but it also makes the trait more flexible and I believe in general that bounds on methods and functions are better compared to bounds on traits and structs. For example, in my use case the export functionality is not needed so this PR would allow me to use `Cow<'a, str>` in fields with no negatives. Let me know what you think.

### DIFF
--- a/ts-rs/src/lib.rs
+++ b/ts-rs/src/lib.rs
@@ -235,7 +235,7 @@ mod export;
 ///
 /// - `#[ts(skip)]`:  
 ///   Skip this variant  
-pub trait TS: 'static {
+pub trait TS {
     const EXPORT_TO: Option<&'static str> = None;
 
     /// Declaration of this type, e.g. `interface User { user_id: number, ... }`.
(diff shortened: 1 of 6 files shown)


ts-rs/src/lib.rs
```diff
@@ -584,7 +624,7 @@ impl_primitives! {
     u64, i64, NonZeroU64, NonZeroI64,
     u128, i128, NonZeroU128, NonZeroI128 => "bigint",
     bool => "boolean",
-    char, Path, PathBuf, String, &'static str,
+    char, Path, PathBuf, String, str,
```
NyxCode, line 627: love being actually able to implement it directly for `str` :partying_face:

NyxCode:
looks like a great change! I'll merge it & maybe write some more tests, but everything looks good to me. Thanks!

PR #139, opened 2023-02-03
Suppress warnings feature

rcvfdmark:
Install non-default feature "suppress-warnings" to suppress all the compile-time messages about ts-rs being unable to parse serde annotations.


NyxCode:
Thanks for the PR. This is supported on the latest master with the `no-serde-warnings` feature.

PR #141, opened 2023-02-14
Fix type struct with zero fields

Maan2003:
types of `struct Foo();` and `struct Foo{}` are incorrect. Type should be `[]` and `{}` for compat with serde json.

This is just a simple patch, I will cleanup it up and fix the tests if approved.


NyxCode:
fixed in #161

Issue #105, opened 2022-05-29
How to use #[ts(export_to = "")]?


yujong-lee:
When I use `#[ts(export)]`, bindings are generated in `bindings` folder. (Same level with `src`).

Now I want to change export path to `pkg/bindings` but I don't know how. I tried something like `#[ts(export_to = "pkg")]`, `#[ts(export_to = "../pkg")]`, but this seems to not work.

Can you please explain to me what right way to use `#[ts(export_to = "")] please?`

Thank you.

NyxCode:
Hey @yujong-lee,
if you want to export to a specific file, you can do `#[ts(export_to = "path/to/file")]`.  
If you want to export to a directory, you can do `#[ts(export_to = "path/to/directory/")]` (notice the trailing `/`).  
I'll clarify this further in the docs.

yujong-lee:
@NyxCode  

```bash
.
├── Cargo.toml
├── bindings
├── pkg
│   └── bindings
├── src
│   ├── graph
│   │   ├── types.rs
```

I have many `#[ts()]` macros is in `src/graph/types.rs` like below.

```rs
#[derive(Serialize, Clone, TS)]
#[ts(export)]
pub struct Node {
}
```

Default output is `bindings`. I want `pkg/bindings` to be output path.

Then isn't  `#[ts(export_to = "pkg/bindings/")]` correct? 

Currently, it doesn't works for me.

NyxCode:
You need `#[ts(export)]`, as well as `#[ts(export_to = "pkg/bindings/")]`:  
```rust
#[derive(Serialize, Clone, TS)]
#[ts(export, export_to = "pkg/bindings/")]
pub struct Node {
}
```
`export` generates a test which does the exporting, `export_to` sets the output path.

yujong-lee:
@NyxCode Now I got it. Thank you so much. 👍

PR #145, opened 2023-03-07
Add doc note about the necessity to use `export`

eirikblekesaune:
We found ourselves not being able to export to a specific directory today. We found the answer to the problem in this issue: https://github.com/Aleph-Alpha/ts-rs/issues/105#issuecomment-1140435797 Figured it would be a good idea to make a note of this for future users.


NyxCode:
thanks!

PR #146, opened 2023-03-20
Don't warn about `#[serde(with = "...")]`

lynn:
I had some code like this:

```rs
use chrono::serde::ts_seconds;
use chrono::DateTime;
use serde::{Deserialize, Serialize};
use ts_rs::TS;

#[derive(Serialize, Deserialize, TS)]
#[ts(export)]
pub struct Test {
    #[ts(type = "number")]
    #[serde(with = "ts_seconds")]
    some_time: DateTime<Utc>,
}
```

ts-rs already handles this perfectly, as long as `number` really is the type that `ts_seconds` transforms the field to/from. However, it issues a warning about unrecognized attribute `#[serde(with)]`. So I wrote this patch that parses this attribute and silences the warning.

### DIFF
--- a/macros/src/attr/field.rs
+++ b/macros/src/attr/field.rs
@@ -74,5 +74,9 @@ impl_parse! {
                 parse_assign_str(input)?;
             }
         },
+        // parse #[serde(with = "...")] to not emit a warning
+        "with" => {
+            parse_assign_str(input)?;
+        },
     }
 }


escritorio-gustavo:
This has been implemented in #185 through the `quiet-serde` flag, which will disable all serde related warnings

NyxCode:
I think we still might want to eagerly parse the `with` attribute, and only emit a warning when the field does not have a `#[ts(type = "..")]` override.

escritorio-gustavo:
> and only emit a warning when the field does not have a #[ts(type = "..")] override.

Is it always the case that `with` requires `type`? If so, we could emit a `compile_error!` to make this mistake impossible

escritorio-gustavo:
Implemented in #280

PR #154, opened 2023-06-30
add url support

laptou:


NyxCode:
thanks!

gustavo-shigueo:
> Why not use `URL` for this?

Serializing a rust `url::Url` with `serde` into JS produces a `string`, not an `URL` object

PR #155, opened 2023-06-30
Implement time Duration

tlbdk:


NyxCode:
thanks

NyxCode:
I will revert this change.
The goal is that all types supported out-of-the-box are compatible with how serde_json serializes them. `std::time::Duration` is, by default, serialized as `{secs: .., nanos: ..}`.

Issue #158, opened 2023-07-24
Useless use of `vec!`


googleben:
Clippy complains about [this line](https://github.com/Aleph-Alpha/ts-rs/blob/b4ba8b81fd8833296e99285eae7608864c52e51e/macros/src/deps.rs#L41) in the codebase, saying it's a useless use of `vec!` (which it is). This can be pretty easily fixed like so:

```Rust
fn to_tokens(&self, tokens: &mut TokenStream) {
    let dependencies = &self.0;
    let num_deps = dependencies.len();
    tokens.extend(quote! {
        {
            let mut dependencies = Vec::with_capacity(#num_deps);
            #( #dependencies )*
            dependencies
        }
    })
}
```

Just add a line and change `vec!` to a `Vec::with_capacity` call, fixing the warning and slightly improving performance. There may be a slightly better/cleaner way, but I'm not that well-versed in proc macros, so this at least works and is faster.

I found this issue through [Lemmy](https://github.com/LemmyNet/lemmy), which uses ts_ts v6.2.1 - there are literally hundreds of these warnings in the codebase because of how many structs derive `TS` - and I would really appreciate if this fix was applied to both the main branch and a temporary one so we could get a v6.2.2 or something.

Please let me know if there's anything I can do to help make this happen.

NyxCode:
fixed on master, and in 0.7 to be released soon

PR #159, opened 2023-07-30
Fix useless_vec lints

rye:
Resolves GH-158.

Beta clippy recently started failing on a downstream consumer ([issue](https://github.com/frog-pond/ccc-server-next/issues/83)) of this crate. This was because of a number of calls to `vec!` that were picked up by clippy. This lint only fires, at time of writing, in beta (stable does not have the change that introduced it, and nightly has since changed to ignore this lint in macros...), but I figured I might as well clean it up and take care of GH-158. This will, assuming this change gets released quickly, help my fellow `#![deny(clippy::all)]` folks who use this crate.

I've also removed some trailing whitespace in `ts-rs/src/lib.rs`. I did this in a separate commit, so d4dfefe can be reviewed on its own.

### DIFF
--- a/ts-rs/src/lib.rs
+++ b/ts-rs/src/lib.rs
@@ -343,10 +343,10 @@ macro_rules! impl_tuples {
     ( impl $($i:ident),* ) => {
         impl<$($i: TS),*> TS for ($($i,)*) {
             fn name() -> String {
-                format!("[{}]", vec![$($i::name()),*].join(", "))
+                format!("[{}]", [$($i::name()),*].join(", "))
             }
             fn inline() -> String {
-                format!("[{}]", vec![ $($i::inline()),* ].join(", "))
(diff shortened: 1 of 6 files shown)


rye:
I confirmed that, by swapping over to my fork and rebuilding the consuming application, that the `useless_vec` lints go away.

ts-rs/src/lib.rs
```diff
@@ -70,31 +70,31 @@
 //! - type aliases must not alias generic types (#70)
 //!
 //! ## cargo features
-//! - `serde-compat` (default)  
+//! - `serde-compat` (default)
```
NyxCode, line 73: Seems like something removed the double whitespace at the end of lines in the docs. i think this will screw up line breaks in rustdoc

NyxCode:
merged manually, thanks!

PR #164, opened 2023-08-15
Check for empty tuple in `format_type`

Heliozoa:
Empty tuple fields `()` are currently identified as `never[]` but are serialized into `null`. This PR fixes this by checking for an empty tuple at `format_type`. This can come up in generics where `T = ()`.


NyxCode:
I cant believe that wasn't already covered by the test suite. Thanks!!

PR #172, opened 2023-10-11
Feat: Added Result<T, E> support

### LINKED ISSUE #171: Support the Result<V, E> type
I suggest that a default implementation is addded for the Result<V, E> type of the standard library

KronsyC:
Resolves #171

### DIFF
--- a/ts-rs/src/lib.rs
+++ b/ts-rs/src/lib.rs
@@ -459,6 +459,23 @@ impl<T: TS> TS for Option<T> {
     }
 }
 
+impl<T : TS, E : TS> TS for Result<T, E>{
+    fn name() -> String{
+        unreachable!();
+    }
+    fn transparent() -> bool {
+        return true;
+    }
+    fn dependencies() -> Vec<Dependency>
+    where
+        Self: 'static {
+        [Dependency::from_ty::<T>()].into_iter().flatten().collect()
+    }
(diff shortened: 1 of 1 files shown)


ts-rs/src/lib.rs
```diff
@@ -459,6 +459,23 @@ impl<T: TS> TS for Option<T> {
...
+        unreachable!();
+    }
+    fn transparent() -> bool {
+        return true;
+    }
+    fn dependencies() -> Vec<Dependency>
+    where
+        Self: 'static {
+        [Dependency::from_ty::<T>()].into_iter().flatten().collect()
+    }
```
axos88, line 473: I think the dependency on E is missing

ts-rs/src/lib.rs
```diff
@@ -459,6 +459,23 @@ impl<T: TS> TS for Option<T> {
...
+    fn inline() -> String {
+        format!("{{ ok : {} }} | {{ err : {} }}", T::inline(), E::inline())
+    }
```
axos88, line 476: This should be capital cased though,  shouldn't it?

KronsyC:
I have addressed both of your issues in the most recent commit

NyxCode:
Thanks @KronsyC for the PR, and thanks @axos88 for taking a look at it. Really appreciate it!

PR #173, opened 2023-10-14
Made the TS implementation for hashmap and hashset generic over all hashers

VictorKoenders:
By default the hasher is [RandomState](https://doc.rust-lang.org/stable/std/collections/hash_map/struct.RandomState.html) but developers can override this. This PR makes ts-rs usable with e.g. https://docs.rs/rustc-hash/latest/rustc_hash/.

### DIFF
--- a/ts-rs/src/lib.rs
+++ b/ts-rs/src/lib.rs
@@ -104,7 +104,7 @@
 //!   Implement `TS` for `OrderedFloat` from ordered_float
 //!
 //! - `heapless-impl`  
-//! 
+//!
 //!   Implement `TS` for `Vec` from heapless
 //!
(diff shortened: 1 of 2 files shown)


VictorKoenders:
Added tests for `HashMap`, `HashSet` and examples on how to use custom hashers.

It seems like you can't use `HashMap<K, V, CustomHasher>` directly because the `TS` derive requires that all type arguments extend from `TS`. However with a type alias it works perfectly fine.

NyxCode:
Great! Don't see any problems, here, so let's get this merged. Thanks @VictorKoenders for the PR, and thanks @escritorio-gustavo for taking a look!

Issue #113, opened 2022-08-12
Workaround for foreign types


hut8:
Thanks so much for such a great library. It's very useful and I really appreciate the effort that's been put in!

Much like in #82, many of my structs have foreign types. Two of the ones I have that are probably quite common are `url::Url` (https://docs.rs/url/latest/url/) and `compact_str::CompactString` (https://docs.rs/compact_str/latest/compact_str/). I am not a rust expert so if my understanding of the issue is flawed, please forgive me 😄 

I believe that because of the orphan rule, a derive macro cannot generate code for a type defined in a different crate. Serde has a workaround: https://serde.rs/remote-derive.html -- I could attempt to implement this for ts-rs if you think this is the right way forward.

For now I'm just generating a `string` via `ts(type = 'string')` which is fine but with some types misses quite a bit.

NyxCode:
Hey, thanks for the kind words!

You're completely right that the orphan rule is the problem here.
For now, the solution was just to add a *lot* of optional cargo features to ts-rs. When enabled, these then implement `TS` for the corresponding crate.

I'm not against having something comparable to serde's remote derive. For me, having features like `serde-compat` for the most common crates and overriding types with `#[ts(type = '..')]` has been good enough though.

For this library, it would probably just require one attribute, maybe `#[ts(as = some::other::Type)]`. 
If questions come up while implementing this, feel free to ask, happy to help.

ranmocy:
While `as` attribute seems could solve the issue, it would be more generic to support Serde [with/remote](https://serde.rs/remote-derive.html) instead?
```rs
#[derive(Debug, Clone, Serialize, Deserialize, TS)]
#[serde(remote = "ExternalType")]
pub struct ExternalTypeDef {
    #[ts(type = "string[]")]
    pub data: Bytes,
}

#[derive(Debug, Clone, Serialize, Deserialize, TS)]
pub enum MyEnum {
    #[serde(with = "ExternalTypeDef")]
    Some(ExternalType),
    Other(String),
}
```

PR #174, opened 2023-10-17
add "as" derive helper attribute

josema03:
This implements the feature requested in the following issue: https://github.com/Aleph-Alpha/ts-rs/issues/113

Please take a look at it and let me know if I need to make any other change to have it merged.


macros/src/types/enum.rs
```diff
@@ -95,12 +95,26 @@ fn format_variant(
...
             Fields::Unnamed(unnamed) if unnamed.unnamed.len() == 1 => {
-                let FieldAttr { type_override, .. } =
-                    FieldAttr::from_attrs(&unnamed.unnamed[0].attrs)?;
+                let FieldAttr {
+                    type_as,
+                    type_override,
+                    ..
+                } = FieldAttr::from_attrs(&unnamed.unnamed[0].attrs)?;
+
+                if let (Some(_type_as), Some(_type_override)) = (&type_as, &type_override) {
```
escritorio-gustavo, line 104: Since you're just checking if the `Options` are `Some`, the following code is easier to understand:
```rs
if type_as.is_some() && type_override.is_some() {
```
It removes the need for `if let`, tuples and borrowing

macros/src/types/enum.rs
```diff
@@ -120,13 +134,28 @@ fn format_variant(
...
                 Fields::Unnamed(unnamed) if unnamed.unnamed.len() == 1 => {
-                    let FieldAttr { type_override, .. } =
-                        FieldAttr::from_attrs(&unnamed.unnamed[0].attrs)?;
+                    let FieldAttr {
+                        type_as,
+                        type_override,
+                        ..
+                    } = FieldAttr::from_attrs(&unnamed.unnamed[0].attrs)?;
+
+                    if let (Some(_type_as), Some(_type_override)) = (&type_as, &type_override) {
```
escritorio-gustavo, line 143: Use `.is_some()` instead

macros/src/types/enum.rs
```diff
@@ -120,13 +134,28 @@ fn format_variant(
...
+                        syn_err!("`type` is not compatible with `as`")
+                    }
+
+                    let parsed_ty = if let Some(_type_as) = &type_as {
```
escritorio-gustavo, line 147: `_type_as` in not unused, so it should not have an underscore. If you are worried about name collisions and shadowing, give the variable a diferent name

Also, to prevent the move, instead of
```rs
if let Some(type_as) = &type_as {
```

you can probably use
```rs
if let Some(ref type_as) = type_as {
```

I haven't checked this against the compiler, but if it does compile I think you should write it this way

macros/src/types/named.rs
```diff
@@ -75,16 +76,29 @@ fn format_field(
         return Ok(());
     }
 
-    let (ty, optional_annotation) = match optional {
-        true => (extract_option_argument(&field.ty)?, "?"),
-        false => (&field.ty, ""),
+    if let (Some(_type_as), Some(_type_override)) = (&type_as, &type_override) {
```
escritorio-gustavo, line 79: Use `.is_some()`

macros/src/types/named.rs
```diff
@@ -75,16 +76,29 @@ fn format_field(
...
+        syn_err!("`type` is not compatible with `as`")
+    }
+
+    let parsed_ty = if let Some(_type_as) = &type_as {
```
escritorio-gustavo, line 83: Rename the variable to remove the `_` prefix, and if possible use `ref` instead of `&`

macros/src/types/named.rs
```diff
@@ -75,16 +76,29 @@ fn format_field(
...
+    } else {
+        field.ty.clone()
+    };
+
+    let (ty, optional_annotation) = {
+        match optional {
+            true => (extract_option_argument(&parsed_ty)?, "?"),
+            false => (&parsed_ty, ""),
+        }
     };
```
escritorio-gustavo, line 94: Why the extra brackets?

```rs
let (ty, optional_annotation) = match optional {
    true => (extract_option_argument(&parsed_ty)?, "?"),
    false => (&parsed_ty, ""),
};
```

macros/src/types/newtype.rs
```diff
@@ -38,17 +39,26 @@ pub(crate) fn newtype(
         _ => {}
     };
 
-    let inner_ty = &inner.ty;
+    if let (Some(_type_as), Some(_type_override)) = (&type_as, &type_override) {
```
escritorio-gustavo, line 42: Use `.is_some()`

macros/src/types/tuple.rs
```diff
@@ -71,6 +71,15 @@ fn format_field(
     if skip {
         return Ok(());
     }
+    
+    let ty = if let Some(_type_as) = &type_as {
```
escritorio-gustavo, line 75: Rename the variable to remove the `_` prefix, and if possible use `ref` instead of `&`

macros/src/types/newtype.rs
```diff
@@ -38,17 +39,26 @@ pub(crate) fn newtype(
...
+        syn_err!("`type` is not compatible with `as`")
+    }
+
+    let inner_ty = if let Some(_type_as) = &type_as {
```
escritorio-gustavo, line 46: Rename the variable to remove the `_` prefix, and if possible use `ref` instead of `&`

escritorio-gustavo (changes requested):

macros/src/types/enum.rs
escritorio-gustavo, line 104: Since you're just checking if the `Options` are `Some`, the following code is easier to understand:
```rs
if type_as.is_some() && type_override.is_some() {
```
It removes the need for `if let`, tuples and borrowing

macros/src/types/enum.rs
escritorio-gustavo, line 143: Use `.is_some()` instead

macros/src/types/enum.rs
escritorio-gustavo, line 147: `_type_as` in not unused, so it should not have an underscore. If you are worried about name collisions and shadowing, give the variable a diferent name

Also, to prevent the move, instead of
```rs
if let Some(type_as) = &type_as {
```

you can probably use
```rs
if let Some(ref type_as) = type_as {
```

I haven't checked this against the compiler, but if it does compile I think you should write it this way

macros/src/types/named.rs
escritorio-gustavo, line 79: Use `.is_some()`

macros/src/types/named.rs
escritorio-gustavo, line 83: Rename the variable to remove the `_` prefix, and if possible use `ref` instead of `&`

macros/src/types/newtype.rs
escritorio-gustavo, line 42: Use `.is_some()`

macros/src/types/newtype.rs
escritorio-gustavo, line 46: Rename the variable to remove the `_` prefix, and if possible use `ref` instead of `&`

macros/src/types/tuple.rs
escritorio-gustavo, line 75: Rename the variable to remove the `_` prefix, and if possible use `ref` instead of `&`

escritorio-gustavo (changes requested):
Sorry about the mess, the previous comments were not marking the specific lines I was referring to, but the whole set of changes

macros/src/types/enum.rs
```diff
@@ -95,12 +95,26 @@ fn format_variant(
...
+                    type_as,
+                    type_override,
+                    ..
+                } = FieldAttr::from_attrs(&unnamed.unnamed[0].attrs)?;
+
+                if let (Some(_type_as), Some(_type_override)) = (&type_as, &type_override) {
+                    syn_err!("`type` is not compatible with `as`")
+                }
+
+                let parsed_ty = if let Some(_type_as) = &type_as {
```
escritorio-gustavo, line 108: Rename the variable to remove the `_` prefix, and if possible use `ref` instead of `&`

escritorio-gustavo:
I have fixed the merge conflict, but I haven't passed the fix through the compiler, needs to be checked

NyxCode:
I love it! Thanks @josema03, and thank you so much @escritorio-gustavo for putting in the time to clean this up!  
Once it's ready, I think this would be a really cool addition.

escritorio-gustavo:
Hey @NyxCode!
I'm having some trouble merging this with #213, how should these PRs interact in `macros/types/named.rs`

NyxCode:
> Hey @NyxCode! I'm having some trouble merging this with #213, how should these PRs interact in `macros/types/named.rs`
> 
> Edit: I think I got it working (CI passing) but would still like to have you take another look at this file before merging

Sure, lemme take a look at that!

macros/src/types/named.rs
```diff
@@ -92,22 +93,33 @@ fn format_field(
...
+        field.ty.clone()
+    };
+
     let (ty, optional_annotation) = match optional {
         Optional { optional: true, nullable } => {
             let inner_type = extract_option_argument(&field.ty)?; // inner type of the optional
             match nullable {
-                true => (&field.ty, "?"),  // if it's nullable, we keep the original type
+                true => (&parsed_ty, "?"),  // if it's nullable, we keep the original type
                 false => (inner_type, "?"), // if not, we use the Option's inner type
```
escritorio-gustavo, line 111: The thing I'm most confused on how to do is this. This code branch bypasses `type_as`

escritorio-gustavo: Note: I'm talking about line 111, not the lines marked in the diff

NyxCode: Just hit that issue as well. I'll have to think about how it *should* interact, i'm not 100% sure about that yet.

macros/src/types/named.rs
```diff
@@ -92,22 +93,33 @@ fn format_field(
...
+
+    let parsed_ty = if let Some(ref type_as) = type_as {
+        syn::parse_str::<Type>(type_as)?
+    } else {
+        field.ty.clone()
+    };
+
     let (ty, optional_annotation) = match optional {
         Optional { optional: true, nullable } => {
             let inner_type = extract_option_argument(&field.ty)?; // inner type of the optional
```
NyxCode, line 108: I *think* we wanna change this to `let inner_type = extract_option_argument(&parsed_ty)?`.
To get `as` working with `optional` then, you do 
```rust
#[derive(TS)]
struct Outer {
    #[ts(as = "Option<ExternalTypeDef>")]
    #[ts(optional)]
    x: Unsupported
}
```

The alteratives I can think off seem unintuitive, from a users perspective. Like this, the `as` override just.. overrides the field, even if it's an option.

NyxCode: With just that change, I think everything works like a charm:
```rust
#[derive(TS)]
struct Outer {
    #[ts(as = "Option<ExternalTypeDef>")]
    #[ts(optional = nullable, inline)]
    x: Unsupported
}
```
```ts
type Outer = { x?: { a: number, b: number, c: number, } | null, }
```

NyxCode: I've got a commit ready if you think this makes sense @escritorio-gustavo

escritorio-gustavo: I think I get it, but how does `#[ts(optional, as = "...")]` (without nullable) behave? This codepath was ignoring `as`

escritorio-gustavo: Nevermind I just realized `inner_type` is the variable used in the codepath I mentioned

escritorio-gustavo: > I've got a commit ready if you think this makes sense @escritorio-gustavo

Seems good to me!

NyxCode (approved):

NyxCode:
From my side, this is ready to merge! 
@escritorio-gustavo Can you take a last look & merge this? <3

escritorio-gustavo (approved):
LGTM!

PR #176, opened 2023-11-03
feat: impl for semver::Version

tymmesyde:


escritorio-gustavo:
Hey, can you add some tests and fix the merge conflicts?

ts-rs/tests/semver.rs
```diff
@@ -0,0 +1,19 @@
+#![allow(deprecated)]
+#![cfg(feature = "chrono-impl")]
```
escritorio-gustavo, line 2: I think this is supposed to be `#![cfg(feature = "semver-impl")]` right?

ts-rs/tests/semver.rs
```diff
@@ -0,0 +1,19 @@
+#![allow(deprecated)]
```
escritorio-gustavo, line 1: Just curious, what is deprecated in this file?

escritorio-gustavo (commented):
I am confused about the attributes on your test file

ts-rs/tests/semver.rs
tymmesyde: Yeah my bad

ts-rs/tests/semver.rs
tymmesyde: I just forgot to remove this line sorry

escritorio-gustavo (approved):
LGTM, thanks for the PR!

PR #178, opened 2023-11-09
Add feature flags for `schemars` and `serde_json`

jessfraz:
Thanks so much for this awesome lib.

We needed these so we forked it, but would love to push back upstream!


simonsparks:
Hi @jessfraz, I just noticed this PR and it looks really useful for my current problem with serde_json::Value.
I was wondering how hard you think it would be to also support serde_json::Map in this PR?

NyxCode:
Thanks for the PR. However, I see a few problems with this:
- The newline in the docs were altered
- `serde_json::Value` is missing arrays and objects. Maybe `any` is the most appropriate type here, i'm not sure.

I'll close this for now. In the meantime, I'd encourage you to use a type override (`#[ts(type = "..")]`) on the fields.

jessfraz:
I didnt realize i could just use #[ts(type = "..")] as an override thats so much easier thank you!

PR #180, opened 2023-11-29
Added `serde(skip)` functionality inside Enum Variants

timephy:
## Observation

I wanted to skip serializing an enum error variant in a project, similar to this:

```rust
#[derive(Debug, ::thiserror::Error, serde::Serialize, ts_rs::TS)]
#[ts(export)]
#[serde(tag = "type", content = "content")]
pub enum Error {
    #[error("Account with id ({0}) not found")]
    NotFound(AccountId),
    #[error(transparent)]
    Any(
        #[from]
        #[serde(skip)]
        ::anyhow::Error,
    ),
}
```

This produced the following error: "`skip` is not applicable to newtype fields".

## Problem

I wanted to skip serializing the field, to not send it to the client, as I want to keep `Any` errors on the server and treat them as an anonymous `InternalServerError` to the client.

However, I want to keep the internal error data for logging, I do not want to "delete" the error data.

It seemed like I needed a copy of my `Error` and omit the `Any` variants value (`anyhow::Error`), and then create a `From`/`Into` implementation from the `Error` with the `Any` field into the `Error` without the `Any` field.

## Solution

In this pull request I have modified the `ts-rs` code so that it generates the types in the same way that `serde` serializes Enums in these conditions.
This is achieved by treating an Enum Variant with exactly one Unnamed field (which was already done by the code before in some places) and if that field has a `skip` attribute, then treat that Enum Variant as a Unit instead of a Newtype.

## Tests

I implemented tests to guarantee the behaviour for all 4 types of Enum Tags (Untagged, Externally, Adjacently, Internally) both for `Named` and `Unnamed` field variants.


NyxCode:
thanks for the PR, really appreciate it!

NyxCode:
I'll merge this as-is for now, but I'll try to get rid of the repetition of `if skip { .. } else { .. }` before the next release.

PR #181, opened 2023-12-10
fix: Include borrowed types' type arguments in typescript

aatukaj:
Borrowed types' type arguments weren't previously being included in the typescript export.
e.g.
```rust
struct A<'a> {
    b: &'a Vec<usize>,
}
```
Would be exported as: 
```typescript
interface A { b: Array, }
```

With this pr it's exported correctly as:
```typescript
interface A { b: Array<number>, }
```


NyxCode:
Thanks for the PR!

PR #183, opened 2023-12-12
Implement TS trait for slices

escritorio-gustavo:
Use `impl_shadow!` macro to implement `TS` for slices of type `T: TS`.

This allows for structs like this: 
```rust
#[derive(ts_rs::TS)]
#[export, export_to = "./bindings/"]
struct Foo {
    arr: Box<[u8]>
}
```

This is great for saving memory when dealing with immutable collections of unknown size, since there is no `capacity` to store like there is with `Vec<T>`

### DIFF
--- a/ts-rs/Cargo.toml
+++ b/ts-rs/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ts-rs"
-version = "7.0.0"
+version = "7.1.0"
 authors = ["Moritz Bischof <moritz.bischof1@gmail.com>"]
 edition = "2021"
 license = "MIT"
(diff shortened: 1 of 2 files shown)


NyxCode:
Thanks for the PR. Could you add a test & remove the version bump?

escritorio-gustavo:
Sure thing, I was just adding a fix to the derive macro.

escritorio-gustavo:
Are you sure about the version bump removal though? You won't be able to push the update to crates.io due to the version number being the same, and it is a new feature, which normally would require a bump to the minor version to conform to semver

escritorio-gustavo:
Done!

NyxCode:
@escritorio-gustavo yeah, thanks for removing it. I don't publish every PR as a new release, so i'd like to do that separately. It'd also be a backwards-compatible addition, so if we'd publish this PR as a new version, i think it should be 7.0.1.

PR #184, opened 2023-12-21
Make sure that `rename_all = "camelCase"` works like in serde

### LINKED ISSUE #165: Mismatch between `ts-rs`'s idea of `camelCase` and `serde`'s
I have a type like this:
```rs
#[derive(Debug, Serialize, TS)]
#[serde(rename_all = "camelCase")]
#[ts(export)]
pub struct ItemImageList {
    pub id: u32,
    pub crc32c_hash: u32,
}
```
The generated camelCase field name from `serde` for the field `crc32c_hash` is `crc32cHash`. `ts-rs`, on the other hand, decides to generate the following TS file:
```ts

export interface ItemImageList { id: number, crc32CHash: number, }
```

This obviously leads to problems when trying to use the `ItemImageList` interface in TS to read JSON serialized by serde.

The solution for me for the time being, has been to slap `#[ts(rename = "crc32cHash")]` on the field. But the question is, should there actually be a discrepancy here?

escritorio-gustavo:
Fixes #165 

This copies over serde's implementation of Pascal and camelCase to make sure both crates work the same way

edit: typos

### DIFF
--- a/macros/src/attr/mod.rs
+++ b/macros/src/attr/mod.rs
@@ -32,9 +32,29 @@ impl Inflection {
         match self {
             Inflection::Lower => string.to_lowercase(),
             Inflection::Upper => string.to_uppercase(),
-            Inflection::Camel => string.to_camel_case(),
+            Inflection::Camel => {
+                let pascal = Inflection::apply(Inflection::Pascal, string);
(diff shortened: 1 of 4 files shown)


NyxCode:
Thanks for the PR!

PR #185, opened 2023-12-21
Add `quiet-serde` feature flag

### LINKED ISSUE #151: prevent warning for deserialize_with- Require Option<T> value, where null == None
I got cases where i want the values to be specified and null indicates that it's not used.
To do this i got something like:

```rust

#[derive(TS, Deserialize, Serialize, Debug, Clone)]
#[ts(export)]
pub struct Foo{
    pub id: i32,
    #[serde(deserialize_with = "require_value_or_null")]
    pub value: Option<String>,
}
pub fn require_value_or_null<'de, D, T>(deserializer: D) -> Result<Option<T>, D::Error>
where
    D: Deserializer<'de>,
    for<'a> T: Deserialize<'a>,
{
    #[derive(Deserialize, Debug)]
    #[serde(untagged)]
    enum Aux<T> {
        T(T),
        Null,
    }
    match Deserialize::deserialize(deserializer)? {
        Aux::T(t) => Ok(Some(t)),
        Aux::Null => Ok(None),
    }
}
```
This works but as `deserialize_with` isn't supported by TS it will throw a warning.
Is there a way to achieve this without that?

### LINKED ISSUE #170: Request: silently ignore werde alias
My compilation is filled with warnings along these lines:

```
warning: failed to parse serde attribute
  |
  | #[serde(alias = "URLs")]
  |
  = note: ts-rs failed to parse this attribute. It will be ignored.
```

Since this is a well-known serde field attribute (https://serde.rs/field-attrs.html), even though ts-rs does not act on it, it should not emit warning cluttering compilation. Thanks for considering.

escritorio-gustavo:
Add a feature flag to make warnings caused by serde attributes silent

Fixes #170 
Fixes #151

### DIFF
--- a/macros/Cargo.toml
+++ b/macros/Cargo.toml
@@ -10,6 +10,7 @@ repository = "https://github.com/Aleph-Alpha/ts-rs"
 
 [features]
 serde-compat = ["termcolor"]
+quiet-serde = []
 
 [lib]
 proc-macro = true
--- a/macros/src/utils.rs
+++ b/macros/src/utils.rs
@@ -124,6 +124,7 @@ mod warning {
 
     // Sadly, it is impossible to raise a warning in a proc macro.
     // This function prints a message which looks like a compiler warning.
+    #[allow(unused)]
     pub fn print_warning(
         title: impl Display,
         content: impl Display,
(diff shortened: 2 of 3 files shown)


NyxCode:
Thanks for the PR! I'll merge this as-is, and add a note in the docs for it.

Issue #52, opened 2021-11-09
export cargo docs as tsdoc


dakom:
this might be too big of an ask... but would be cool =)

NyxCode:
actually, doc comments (`/// something`) are just syntactic sugar for `#[doc]`: https://doc.rust-lang.org/rustdoc/the-doc-attribute.html
If you feel like it, feel free to open a PR. Always happy to help if you've got questions.

riellyd:
Are there any plans to retain any kind of comments?  It helps the front end dev workflow when you don't have to look up the rust type to read any comments about usage, particularly when the files do not site adjacent to one another.  Thanks!

NyxCode:
> Are there any plans to retain any kind of comments? It helps the front end dev workflow when you don't have to look up the rust type to read any comments about usage, particularly when the files do not site adjacent to one another. Thanks!

I'd like to implement this feature. I'm currently a bit busy, so If you feel like it, open a PR.  
We can only retain doc comments, since normal comments are stripped before proc macros are run.

riellyd:
Would love to help, but sadly I am in the same boat as you - rather busy right now. Also, have only done basic proc macros a while ago so would be slow going for me.  Perhaps when things are a bit quieter I'll take a stab if nobody else gets round to doing it, would be a nice feature to have for bigger projects.

escritorio-gustavo:
Closing this issue in favor of tracking it through #188

Issue #86, opened 2022-02-13
Export enum variants as distinct types


tbillington:
Hey I'm really enjoying using this library so far :)

Have you thought about providing the option to export enum variants as distinct types?

Say I have

```rust
#[derive(Serialize, Deserialize, Debug, TS)]
#[ts(export)]
pub enum Cmd {
    Process(u32),
    Stop(String),
}
```

it will produce

```typescript
export type ServerCommand = 
    | { Process: number }
    | { Stop: string }
```

However, while TS can infer if I'm correctly constructing a `ServerCommand`, it would be helpful to be able to type specific variants.

So the result would end up something like this...

```typescript
export type ServerCommand_Process = { Process: number };
export type ServerCommand_Stop = { Stop: string };
export type ServerCommand = 
    | ServerCommand_Process
    | ServerCommand_Stop;
```

NyxCode:
I explored a simple approach to implementing this:

Example:
```rs
#[derive(TS)]
#[ts(export, split, export_to = "tests-out/split_enum/")]
enum X {
    A(i32),
    B { x: i32, b: i32 },
}
```

In `enum_def`, we generate a type for each variant:
```rs
#[derive(ts_rs::TS)]
struct X_A(i32);

#[derive(ts_rs::TS)]
struct X_B {
    x: i32,
    b: i32,
}
```
and recursively call `enum_def` with this modified `EnumItem`:
```rs
#[ts(export, export_to = "tests-out/split_enum/")] 
enum X { A(X_A), B(X_B) }
```

Shearerbeard:
I like this approach as we have a few on our team that would prefer distinct types for rust enums - is there still a possibility of this becoming a feature?

PR #187, opened 2024-01-06
feat: Add functionality to export doc strings on types

timephy:
# Overview

I added functionality (and tests) to export doc strings on exported types.

# Example

```rust
/// Doc comment.
/// Supports new lines.
///
/// Testing
#[derive(TS)]
struct A {
    name: String,
}
```

is exported as

```ts
/**
 * Doc comment.
 * Supports new lines.
 *
 * Testing
 */
export interface A { name: string, }
```

# How

This is achieved by parsing the `Vec<Attribute>` of an the exported type `Item` by filtering for `"doc"` , and then passing that data down to the `export` layer where `generate_decl` prepends the doc string before the export definition in TypeScript.

# Comments

- The only thing I am unsure about is how this interacts with the `type_def` function which is used both for exported types and field types.
- Only works with "normal doc comments", not with `include_str` doc comments or other "doc formats".

# Related

#52
#102


bennyhodl:
Concept ACK. Would love to see this merged

https://github.com/Aleph-Alpha/ts-rs/issues/188

escritorio-gustavo:
Fixes #188 
Just so GitHub tracks the issue with the PR

NyxCode:
Thanks for the PR! 
The general structure of the PR looks good, though there are some minor problems I'd like to resolve before merging this.

escritorio-gustavo:
Hey @NyxCode! What are those minor problems? I can take a look when I have time if you want

NyxCode:
> Hey @NyxCode! What are those minor problems? I can take a look when I have time if you want

oh, seems like i didnt submit the review, sorry.

NyxCode (changes requested):

escritorio-gustavo:
Well this is annoying, locally editing this PR gets me 131 commits behind main, so tests still expect `interface`, but on CI they expect `type`

escritorio-gustavo:
@timephy, can you merge main into your fork? I tried to do it but it just turns into a PR to your fork

escritorio-gustavo:
Nevermind, figured out how to do it!
Sorry, still new to managing PRs

macros/src/types/mod.rs
```diff
@@ -14,26 +18,28 @@ pub(crate) use r#enum::r#enum_def;
...
 
 fn type_def(
     attr: &StructAttr,
+    attrs: &Vec<Attribute>,
     ident: &Ident,
     fields: &Fields,
     generics: &Generics,
 ) -> Result<DerivedTS> {
     let name = attr.rename.clone().unwrap_or_else(|| to_ts_ident(ident));
+    let docs = utils::get_docs_from_attributes(attrs);
```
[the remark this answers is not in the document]
escritorio-gustavo: Also needs to be done for `EnumAttr`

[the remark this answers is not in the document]
escritorio-gustavo: Also renamed the function to parse_docs

macros/src/utils.rs
```diff
@@ -116,6 +116,24 @@ pub fn parse_serde_attrs<'a, A: TryFrom<&'a Attribute, Error = Error>>(
...
+            _ => None,
+        })
+        .map(|attr| match attr.value {
+            Expr::Lit(ExprLit {
+                lit: Lit::Str(ref str),
+                ..
+            }) => Ok(str.value()),
+            _ => syn_err!(attr.span(); "doc attribute with non literal expression found"),
+        })
+        .collect::<Result<Vec<_>>>()
```
escritorio-gustavo, line 134: Managed to get rid of those `panic!`s

macros/src/utils.rs
```diff
@@ -116,6 +116,24 @@ pub fn parse_serde_attrs<'a, A: TryFrom<&'a Attribute, Error = Error>>(
         .into_iter()
 }
 
+/// Return a vector of all lines of doc comments in the given vector of attributes.
+pub fn parse_docs(attrs: &[Attribute]) -> Result<Vec<String>> {
```
escritorio-gustavo, line 120: Always prefer `&[T]` over `&Vec<T>`

ts-rs/src/export.rs
```diff
@@ -249,3 +256,17 @@ where
...
     }
 }
+
+/// Returns an unindented docstring that has a newline at the end if it has content.
+fn format_docs(docs: &str) -> String {
+    match docs.is_empty() {
+        true => "".to_string(),
+        false => {
+            let lines = docs
+                .lines()
```
escritorio-gustavo, line 266: Interestingly, changing docs from `Vec<String>` to `&str` only changed this line (`.iter()` to `.lines()`) in `format_docs`

escritorio-gustavo (commented):
These are the notable changes I made on top of what @NyxCode asked

timephy:
Hi!

First of all thank your for refactoring and fixing problems!

I have changed the code to still generate a `const DOCS = None` field, even if no docs are set on the type.

I use the `.docs()` or now `.DOCS` in code of mine, and I rely on its presence, therefore if the field is simply missing (instead of being None) this would lead to errors.

timephy:
Hi,

First of all, thanks for refactoring the code and solving problems!
Especially that you removed `docs` from all function calls, and moved it into `Attr` structs. - This bothered me as well, but I was not sure about how to solve it better at the time.

Note: I made a commit I then realized was unnecessary, because the `TS` trait already implemented a fallback for the `DOCS` field. Therefore I removed those changes again.

---

I believe it is possible to also add docs for struct fields and enum variants... (#188)
Do we want to implement and include this in this PR or implement it somewhen else in another?

timephy:
Okay, I tried to implement docs for fields+variants as well..

But there were several complications that we need to resolve, I'd like to ask for your help doing that / making the right decisions. I have committed these changes to a separate branch and opened a new PR #228 for that.

I suggest implementing this PR first / implementing the docs for fields/variants in that new PR.

Thanks again for the help/coop!

escritorio-gustavo:
> I believe it is possible to also add docs for struct fields and enum variants

I think struct fields shouldn't be **too** dificult, but we have to make sure to handle `#[ts(flatten)]` correctly, which always complicates things

Enum variants are a whole other beast though. They are represented as a type union of the different variants and there is no convenient way to add docs for specific type union members in TS.

We could implement #86 and document each variant type, but at the end of the day, we'd do `type Enum = VariantA | VariantB`, and when using `Enum` the docs for `VariantA` and `VariantB` would not show up anywhere

escritorio-gustavo:
> Do we want to implement and include this in this PR or implement it somewhen else in another?

Definetly another PR, creating #228 was a good call

> I suggest implementing this PR first / implementing the docs for fields/variants in that new PR.

Agreed, since all the changes requested have already been made, I will merge this PR now

PR #191, opened 2024-01-19
Fix: rename remaining quiet serde feature uses

stegaBOB:
It looks like a couple uses of the old `quiet-serde` feature were missed in #189 when it was renamed. As a result, warnings will still display with the latest version even with the feature on unfortunately.


NyxCode:
Thanks, that's a big blunder. That change must have gotten lost during a merge or something, I specifically recall changing that ^^

Issue #135, opened 2022-12-09
`warning: failed to parse serde attribute #[serde(rename_all = "camelCase")]` on enum struct variant


ctison:
Hello,

I set `#[serde(rename_all = "camelCase")`[^1] on enum struct variants:
```rust
#[derive(Serialize, Deserialize, TS, Clone)]
#[ts(export, export_to = "../src/model/")]
#[serde(rename_all = "camelCase")]
pub enum TaskStatus {
  #[serde(rename_all = "camelCase")]
  Running { started_time: chrono::DateTime<Utc> },
  #[serde(rename_all = "camelCase")]
  Terminated {
    status: i32,
    stdout: String,
    stderr: String,
  },
}
```

[^1]: https://serde.rs/variant-attrs.html#rename_all

But I have the following warnings:

```
warning: failed to parse serde attribute
  | 
  | #[serde(rename_all = "camelCase")]
  | 
  = note: ts-rs failed to parse this attribute. It will be ignored.
```

And the generated Typescript for the enum struct variants' fields are in `snake_case` (eg. `started_time`).

Version: `ts-rs = { version = "6.2.1", features = ["chrono-impl", "uuid-impl"] }`

linulas:
It seems to be an issue with most serde field attributes, I get the same with `#[serde(skip_serializing_if = "Option::is_none")]`, and there's a pull request for fixing the same issue for `[serde(with = "...")]`: #146

NyxCode:
As @escritorio-gustavo mentioned, the `no-serde-warnings` can disable these warnings.  
I think I'd still like to keep this issue open, in case we want to support serde's `rename_all` in the future.

NyxCode:
right now, it's sadly neccessary to rename each field individually with ``#[serde(rename = "..")]` or `#[ts(rename = "..")]`

escritorio-gustavo:
My mistake, due to #165 I thought `#[serde(rename_all = "..")]` was generally supported and this issue was just older than support for it, I didn't notice it didn't apply to enums

escritorio-gustavo:
> right now, it's sadly neccessary to rename each field individually with #[serde(rename = "..")] or #[ts(rename = "..")]

Hey @NyxCode... I have forked the repo to start working on a fix.
I wrote the following test:
```rs
use serde::{Serialize, Deserialize};
use ts_rs::TS;

/// Used enum from issue to test
#[derive(Serialize, Deserialize, TS, Clone)]
[...]

PR #196, opened 2024-01-22
Add test for `rename_all` in enum struct variants

escritorio-gustavo:
Added test requested [here](https://github.com/Aleph-Alpha/ts-rs/issues/135#issuecomment-1902162740)


escritorio-gustavo:
This accidentally inluded commits from a different PR, sorry

PR #198, opened 2024-01-22
Fix bug introduced in #184

escritorio-gustavo:


NyxCode:
💪Awesome, thank you so much!

PR #199, opened 2024-01-22
Improve support for references

escritorio-gustavo:
This PR is meant to improve on #183 as I noticed that `&'a [i32]` would turn into just `Array` int TS rather than `Array<number>`.


NyxCode:
Great work! Thanks!

PR #201, opened 2024-01-23
Change CONTRIBUTING.md to avoid UTF-16 encoding issue

escritorio-gustavo:
The issue found yesterday about README.md having been encoded in UTF-16 is caused by running `cargo readme > file_path` on Windows.

Windows users should instead use `cargo readme -o file_path`


NyxCode:
Thanks! 
PS: I've sent you an email to your proton address, just in case you don't check that regularly.

escritorio-gustavo:
Oh thanks for the heads up, I pretty much never check that lol, I'll go read it.

PR #202, opened 2024-01-23
Allow enum flattening

### LINKED ISSUE #96: Unable to generate flattened enum
I have tried both tagged or untagged enum in struct, not working.

## Code

Tagged:

```rust
use serde::Serialize;
use ts_rs::TS;

#[derive(Serialize, TS)]
#[ts(export)]
struct Foo {
    #[serde(flatten)]
    baz: Bar,
}

#[derive(Serialize, TS)]
#[serde(tag = "tag")]
#[ts(export)]
enum Bar {
    _Baz { a: i32 },
}
```

Untagged:

```rust
use serde::Serialize;
use ts_rs::TS;

#[derive(Serialize, TS)]
#[ts(export)]
struct Foo {
    #[serde(flatten)]
    baz: Bar,
}

#[derive(Serialize, TS)]
#[serde(untagged)]
#[ts(export)]
enum Bar {
    _Baz,
}
```

## Error 

```
---- ts::export_bindings_foo stdout ----
thread 'ts::export_bindings_foo' panicked at 'Bar cannot be flattened', [:REDACTED:]/.cargo/registry/src/github.com-1ecc6299db9ec823/ts-rs-6.1.2/src/lib.rs:243:9
```

## Info
version of `ts-rs`: `6.1.2`
version of `rustc`: `rustc 1.60.0 (7737e0b5c 2022-04-04)`

escritorio-gustavo:
This PR is aimed at allowing the use of `#[flatten]` in enum variants. It's not a very large change, but it does cause some code duplication, by redoing some of what `format_variant` does inside `enum_def`, so I'd like some ideas on how to mitigate that (maybe by changing `format_variant` to return `variant_type`?)

Fixes #96


escritorio-gustavo:
I'm converting this to a draft because I found a massive bug while testing further. My tests only had the enum itself in the struct, but adding other fields didn't work

escritorio-gustavo:
This feature requires changing all uses of `interface` to `type` in order to allow for type intersection (`&` operator)

escritorio-gustavo:
I have changed all the tests to account for this and added a few tests for the flattening of enums.

Although all the tests are passing, I'm still not 100% sure I haven't broken anything

ts-rs/tests/unit.rs
```diff
@@ -5,7 +5,7 @@ use ts_rs::TS;
 struct Unit;
 
 // serde_json serializes this to `{}`.
-// The TS type best describing an empty object is `Record<string, never>`.
+// The TS type best describing an empty object is `{ }`.
```
NyxCode, line 8: I'm curious about this here!
I actually think `Record<string, never>`, while being clunky, is way more appropriate here.
The type `{ }` pretty much allows anything except `null` and `undefined`:
```ts
type Unit = {};

const a: Unit = 123;
const b: Unit = "hey!";
const c: Unit = [1, 2, 3];
const d: Unit = new Map();
const e: Unit = { a: [], b: "hello world" };
const f: Unit = {};
// const g: Unit = null; ERROR
// const h: Unit = undefined; ERROR
```

`Record<string, never>`, on the other hand, actually represents an empty object:
```ts
type Unit = Record<string, never>;

// const a: Unit = 123; ERROR
// const b: Unit = "hey!"; ERROR
// const c: Unit = [1, 2, 3]; ERROR
// const d: Unit = new Map(); ERROR
// const e: Unit = { a: [], b: "hello world" }; ERROR
const f: Unit = {};
// const g: Unit = null; ERROR
// const h: Unit = undefined; ERROR
```

My TS knowledge is wayyyy to basic to explain why `{}` behaves like this, but I guess it does.

NyxCode:
> This feature requires changing all uses of `interface` to `type` in order to allow for type intersection (`&` operator)

I think this is fine. As far as I know, `interfaces` only additionally support [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html). I don't know for sure how common that pattern is, but I suspect not very often.

ts-rs/tests/unit.rs
escritorio-gustavo: > The type `{ }` pretty much allows anything except `null` and `undefined`

This is correct, I just removed it because there seems to be some controversy around the topic [as this thread shows](https://github.com/microsoft/TypeScript/issues/47486#issuecomment-1015542476)

I also did some testing and, although you can't create a `Record<string, never>`, if given one you can read from it and get a `never`, but no error
```ts
declare const test: Record<string, never>
const a = test.foo // no error
   // ^? never
```

escritorio-gustavo:
I have only ever seen declaration merging when I used the `next-auth` library (their examples had it to override some cookies and sessions) as well as people wanting to add to `Window`

ts-rs/tests/unit.rs
NyxCode: Ah, I see. That sucks.
How about `Record<never, never>`?

NyxCode:
Alright! So let's drop this for now, I think that's fine.
I've run a few sanity checks against serde-json, and the output of your flattened enums is consistant with it. Very nice!
Before this PR, I didn't even know it was possible to do that. Pretty awesome stuff.

macros/src/types/named.rs
```diff
@@ -35,18 +35,61 @@ pub(crate) fn named(
...
+                .cloned()
+                .collect::<Box<[_]>>();
+            let other_fields = fields
+                .iter()
+                .filter(|x| !x.contains('{') || match x.find('|') {
+                    Some(i) => &x[i..] == "| null,",
+                    None => true,
+                })
+                .cloned()
+                .collect::<Box<[_]>>();
```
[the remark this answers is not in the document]
NyxCode: Yeah, this seems a bit hacky. Will have to think about this to see if there's a nicer way of doing this.

ts-rs/tests/unit.rs
escritorio-gustavo: As for why `{}` behaves the way it does, take the following code:
```ts
interface Point {
  x: number
  y: number
}

function doSomething(p: Point) {
  return
}

const point = { x: 0, y: 0, z: 0 }
doSomething(point) // No Error
```

In general TS types mean "Object must contain **at least** these properties", not "Object must contain **exactly** this properties", and any type (other than `null` and `undefined`) contain at least zero properties (e.g.: pretty much everything has `.toString`)

escritorio-gustavo: As for `Record<string, never>` it's behaves like "any string may be a key to this object, but no value may be assigned to any of these keys"

escritorio-gustavo:
> Alright! So let's drop this for now, I think that's fine. I've run a few sanity checks against serde-json, and the output of your flattened enums is consistant with it. Very nice! Before this PR, I didn't even know it was possible to do that. Pretty awesome stuff.

Thanks! And I agree, altough everything I tested worked, I just don't trust all the string manipulation I was doing in there

ts-rs/tests/unit.rs
NyxCode: > In general TS types mean "Object must contain at least these properties"

That's how it works in e.g Java, but I don't think that's the entire trough. For example, this here does **not** work in TS. It would work in Java, if `{ a: number, b: number }` was considered a subtype of `{ a: number }`.

```ts
let y: { a: number } = { a: 1, b: 2 };
```

NyxCode:
> > Alright! So let's drop this for now, I think that's fine. I've run a few sanity checks against serde-json, and the output of your flattened enums is consistant with it. Very nice! Before this PR, I didn't even know it was possible to do that. Pretty awesome stuff.
> 
> Thanks! Btw. I think I got somewhat confused. Do you mean drop the PR or the declaration merging?

:D I meant declaration merging! Would love to support flattened enums, I think this effort is great!

escritorio-gustavo:
Great, I might make a separate PR to change `interface`s to `type`s and simplify this one when it's ready

NyxCode:
Awesome, take your time! 

> I also think we need to figure out some better way to identify flattened enum fields, I did a whole lot of string manipulation there that I'm not sure will be able to handle everyone's use cases

I 100% agree! Once you think this is ready, feel free to convert it back to a PR again, then I'll take an other look.  
Let me know as well if you get stuck, I'd definetely be happy to help!

ts-rs/tests/unit.rs
escritorio-gustavo: It seems like TypeScript is more strict about it when it has an object literal. For example, in the code I presented before, the following is invalid:
```ts
interface Point {
  x: number
  y: number
}

function doSomething(p: Point) {
  return
}

doSomething({ x: 0, y: 0, z: 0 })// Error object literal may only specify known properties, and z does not exist in type Point
```

No idea why that is though

escritorio-gustavo:
I think I figured it out. I created a new local branch to start from scratch and managed to get something working by converting all uses of `flatten` into type intersections.

This makes for slightly more redundant TS code but it seems to work. Example with the `flatten` test:
```ts
type Original = { b: { a: number, b: number, c: number, }, d: number, }

type New = { b: { c: number, } & { a: number, b: number, }, d: number, }
//              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ More verbose way to express the same type
```

This pretty much eliminated the string manipulation mess I was previously doing

escritorio-gustavo:
@NyxCode should I close this PR and submit the new branch as a new one or transfer the code over here?

ts-rs/tests/unit.rs
escritorio-gustavo: > Ah, I see. That sucks.
> How about `Record<never, never>`?

This thread got so big I completely missed this comment

`Record<never, never>` gets simplified down to `{}`:
image

escritorio-gustavo: I think there is no way to express that a type can only be assigned an empty object (like `Record<string, never>`) and at the same time get an error reading from it (like `{}`)

escritorio-gustavo:
Btw the extra intersection can probably be solved by adding `.replace(" } & { ", " ")`, which should generate the following:

```ts
type Original = { b: { a: number, b: number, c: number, }, d: number, }

type New = { b: { c: number, a: number, b: number, }, d: number, }
//                ^^^^^^^^^ Less verbose, but the fields will still be out of order (non-flattened fields always come before flattened ones)
```

NyxCode:
> I think I figured it out. I created a new local branch to start from scratch and managed to get something working by converting all uses of `flatten` into type intersections.
> 
> This makes for slightly more redundant TS code but it seems to work. Example with the `flatten` test:
> 
> ```ts
> type Original = { b: { a: number, b: number, c: number, }, d: number, }
> 
> type New = { b: { c: number, } & { a: number, b: number, }, d: number, }
> //              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ More verbose way to express the same type
> ```
> 
> This pretty much eliminated the string manipulation mess I was previously doing

Awesome! Yeah, I don't see any issues with this whatsoever. The generated TS might be a bit more unintuitive that way, but I don't think anyone cares about that.

NyxCode:
> @NyxCode should I close this PR and submit [the new branch](https://github.com/escritorio-gustavo/ts-rs/tree/enum_flattening) as a new one or transfer the code over here?

Whatever you think is best! Generally, I think splitting a change like this into small PRs (doing required refactorings first, then implementing the feature) is a great idea.

escritorio-gustavo:
>  The generated TS might be a bit more unintuitive that way, but I don't think anyone cares about that.

I managed to change that in the [latest commit](https://github.com/escritorio-gustavo/ts-rs/commit/e5929a84883627fadcf19c1b7f54cb01f8268fc3)

PR #203, opened 2024-01-24
Change declarations to use type instead of interface

escritorio-gustavo:
Change TS code generation to use `type` instead of `interface` to make the use of the type intersection (`&`) operator possible


NyxCode:
LGTM! Feel free to merge it.

PR #204, opened 2024-01-24
Remove support for "skip_serializing", "skip_deserializing" and "skip_serializing_if"

### LINKED ISSUE #143: Fields with `skip_deserializing` not getting exported?
Hello!
Thanks for the great crate, I find it very useful!

I'm just wondering if there's a reason why fields marked with `#[serde(skip_deserializing)]` are getting skipped? I saw in [the macro code](https://github.com/Aleph-Alpha/ts-rs/blob/b4ba8b81fd8833296e99285eae7608864c52e51e/macros/src/attr/field.rs#L66) that this is a deliberate decision, but I fail to see the reasoning behind it.

As exporting types are in essence serializing them, I would logically assume that the bindings would still include those fields.

Best regards!

NyxCode:
We plan on removing "support" for these in the next major release.  
Currently, fields annotated with one of these serde annotations are skipped. However, this behavious is incorrect, depending on your usecase. 

Resolves #143

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,5 @@
+# master
+
+- Remove support for "skip_serializing", "skip_serializing_if" and "skip_deserializing".
+    - Initially supporting these by skipping a field was a mistake. If a user wishes to skip a field, they can still
+      annotate it with `#[ts(skip)]`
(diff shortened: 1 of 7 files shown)


escritorio-gustavo (approved):

NyxCode:
great, let's merge it then!

PR #205, opened 2024-01-25
Fix CI badge

escritorio-gustavo:
The current CI badge is showing a `shields` issue rather than the CI status


NyxCode:
Thanks! <3

PR #206, opened 2024-01-25
Enum flattening

### LINKED ISSUE #96: Unable to generate flattened enum

escritorio-gustavo:
Refactored version of #202 created as a separate PR because their histories are not compatible

Fixes #96


NyxCode:
Amazing! The previous PR was super hard to understand (for me, at least :D). This here is now very concise, absolutely love it!

escritorio-gustavo:
It really was, I was still trying to to figure out how to generate `inline_flattened` for enums and it hadn't occurred to me that I could also do type intersections for structs

macros/src/types/enum.rs
```diff
@@ -130,7 +134,7 @@ fn format_variant(
                     "{{ \"{}\": \"{}\", {} }}",
                     #tag,
                     #name,
-                    #inline_flattened
+                    #inline_flattened.trim_matches(&['{', '}', ' '])
```
NyxCode, line 137: I think here, a comment would be usefull - How does `inline_flattened` look like, and why do we need to trim the braces here?

macros/src/types/named.rs
```diff
@@ -57,6 +63,7 @@ pub(crate) fn named(
 // build an expresion which expands to a string, representing a single field of a struct.
 fn format_field(
```
NyxCode, line 64: Same here. Maybe just 1 sentence explaining when the expression ends up in flattened_fields and when in formatted_fields, and why these two vecs are required.

ts-rs/tests/enum_flattening.rs
```diff
@@ -0,0 +1,104 @@
+use serde::Serialize;
```
NyxCode, line 1: We're getting some unused import warnings here.  
Also, we're not testing `#[ts(flatten)]`, which should be equivalent do `#[serde(flatten)]` with compat-mode on. 
I think with a few `#[cfg(..)]`s and `#[cfg_attr(..)]`s, we can make the test work with both compat-mode on and off as-is.

NyxCode (changes requested):

ts-rs/tests/enum_flattening.rs
escritorio-gustavo: Yeah, I'm adding those now, but the only test that will work without `serde-compat` is `internally_tagged` as all the others require `#[serde(tag = "...")]` or `#[serde(untagged)]`

macros/src/types/named.rs
escritorio-gustavo: Done!

macros/src/types/enum.rs
escritorio-gustavo: Done!

escritorio-gustavo: Basically, using [serde's example](https://serde.rs/enum-representations.html#internally-tagged), a type that should turn into this:
```ts
{"type": "Request", id: string, method: string, params: { /*...*/ } }
```

would turn into this
```ts
{"type": "Request", { id: string, method: string, params: { /*...*/ } } }
```

NyxCode (approved):

NyxCode:
Awesomee, thanks for the comments! From my side, I think this is ready for a merge.
Great work, i'm very happy to see support for this. Also, thanks for cleaning up the jungle of open issues & PRs!

PR #208, opened 2024-01-26
Convert fixed size Rust array to fixed size TS tuple

### LINKED ISSUE #137: Allow emitting arrays as TS tuples
```rust
pub(crate) struct Scene {
    color: [f32; 4],
}
```
currently becomes an array of uncertain length,
```typescript
export interface Scene { 
  color: Array<number>;
}
```

It would be useful to be able to mark this e.g.

```rust
pub(crate) struct Scene {
    #[ts(tuple)]
    color: [f32; 4],
}
```
to emit it as
```typescript
export interface Scene { 
  color: [number, number, number, number];
}
```

I tried my hand at implementing this myself, but got lost in macros...

escritorio-gustavo:
Fixes #137


escritorio-gustavo:
Sorry, I included other commits by accident

PR #209, opened 2024-01-26
Convert fixed size Rust array to fixed size TS tuple

### LINKED ISSUE #137: Allow emitting arrays as TS tuples

escritorio-gustavo:
Fixes #137

### DIFF
--- a/ts-rs/src/lib.rs
+++ b/ts-rs/src/lib.rs
@@ -515,7 +515,7 @@ impl<T: TS> TS for Vec<T> {
 }
 
 // Arrays longer than this limit will be emmited as Array<T>
-const ARRAY_TUPLE_LIMIT: usize = 100; // Temp value
+const ARRAY_TUPLE_LIMIT: usize = 128; // Temp value
 impl<T: TS, const N: usize> TS for [T; N] {
     fn name() -> String {
         if N > ARRAY_TUPLE_LIMIT {
--- a/ts-rs/src/lib.rs
+++ b/ts-rs/src/lib.rs
@@ -515,7 +515,7 @@ impl<T: TS> TS for Vec<T> {
 }
 
 // Arrays longer than this limit will be emmited as Array<T>
-const ARRAY_TUPLE_LIMIT: usize = 128; // Temp value
+const ARRAY_TUPLE_LIMIT: usize = 128;
 impl<T: TS, const N: usize> TS for [T; N] {
     fn name() -> String {
         if N > ARRAY_TUPLE_LIMIT {
--- a/ts-rs/src/lib.rs
+++ b/ts-rs/src/lib.rs
@@ -521,7 +521,7 @@ impl<T: TS> TS for Vec<T> {
     }
 }
 
-impl<K: TS, V: TS> TS for HashMap<K, V> {
+impl<K: TS, V: TS, H> TS for HashMap<K, V, H> {
     fn name() -> String {
         "Record".to_owned()
     }
@@ -610,7 +610,7 @@ impl<I: TS> TS for RangeInclusive<I> {
 }
 
 impl_shadow!(as T: impl<'a, T: TS + ?Sized> TS for &T);
-impl_shadow!(as Vec<T>: impl<T: TS> TS for HashSet<T>);
+impl_shadow!(as Vec<T>: impl<T: TS, H> TS for HashSet<T, H>);
 impl_shadow!(as Vec<T>: impl<T: TS> TS for BTreeSet<T>);
 impl_shadow!(as HashMap<K, V>: impl<K: TS, V: TS> TS for BTreeMap<K, V>);
 impl_shadow!(as Vec<T>: impl<T: TS, const N: usize> TS for [T; N]);
--- a/ts-rs/src/lib.rs
+++ b/ts-rs/src/lib.rs
@@ -511,26 +511,44 @@ impl<T: TS> TS for Vec<T> {
     }
 }
 
+// Arrays longer than this limit will be emmited as Array<T>
+const ARRAY_TUPLE_LIMIT: usize = 100; // Temp value
 impl<T: TS, const N: usize> TS for [T; N] {
     fn name() -> String {
-        format!(
-            "[{}]",
-            (0..N).map(|_| T::name()).collect::<Box<[_]>>().join(", ")
-        )
+        if N > ARRAY_TUPLE_LIMIT {
+            return Vec::<T>::name()
+        }
+
+        "[]".to_owned()
     }
 
     fn name_with_type_args(args: Vec<String>) -> String {
+        if N > ARRAY_TUPLE_LIMIT {
+            return Vec::<T>::name_with_type_args(args);
+        }
+
         assert_eq!(
             args.len(),
             1,
-            "called Vec::name_with_type_args with {} args",
+            "called [T; N]::name_with_type_args with {} args",
             args.len()
         );
-        format!("Array<{}>", args[0])
+
+        format!(
+            "[{}]",
+            (0..N).map(|_| args[0].clone()).collect::<Box<[_]>>().join(", ")
+        )
     }
 
     fn inline() -> String {
-        Self::name()
+        if N > ARRAY_TUPLE_LIMIT {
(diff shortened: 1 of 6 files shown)


escritorio-gustavo:
By the way, @NyxCode, if this feature is not something you want, feel free to close the PR and I'll mark the corresponding issue as not planned

NyxCode:
Cool! A couple of points though:

- I feel like we'd definetely need a limit here. Just imagine 
  ```rust
  struct RgbIcon {
      data: [u8; 256 * 256 * 3];
  }
  ``` 
- Why was it necessary to alter the `impl TS for Vec`? Don't quite understand that one yet.

NyxCode:
- We don't implement `TS` for all tuples, so this would fail for large arrays, right?
- Could we alternatively just alter the impl for `[T; N]`?
  (`impl_shadow!(as Vec<T>: impl<T: TS, const N: usize> TS for [T; N]);`)

escritorio-gustavo:
> I feel like we'd definetely need a limit here. Just imagine 
>  ```rust
>  struct RgbIcon {
>      data: [u8; 256 * 256 * 3];
>  }
>  ``` 

You're right, that would be a **massive** TS file! What do you think a sensible limit would be?

> We don't implement TS for all tuples

I didn't realize that was the case, I saw they were converted into a struct with unnamed fields and just assumed they worked for every size, since I couldn't find a limit in `macros/src/types/tuple.rs`.

> Could we alternatively just alter the impl for [T; N]?

Do you mean keep the changes in [`ts-rs/src/lib.rs`](https://github.com/Aleph-Alpha/ts-rs/pull/209/files#diff-7962a4b0d656b7969ae8d7a967fa9c774f84ac70c76383a3c2ca10cbcbede4ec) and discard the ones in [`macros/src/types/generics.rs`](https://github.com/Aleph-Alpha/ts-rs/pull/209/files#diff-30ff1d0665a8474808fa70cb7d5471b77d797d2a4cfcb2e29c41ff6157bb6bba)?
That makes this change not affect struct fields unless they are `#[ts(inline)]`d

escritorio-gustavo:
> Do you mean keep the changes in [ts-rs/src/lib.rs](https://github.com/Aleph-Alpha/ts-rs/pull/209/files#diff-7962a4b0d656b7969ae8d7a967fa9c774f84ac70c76383a3c2ca10cbcbede4ec) and discard the ones in [macros/src/types/generics.rs](https://github.com/Aleph-Alpha/ts-rs/pull/209/files#diff-30ff1d0665a8474808fa70cb7d5471b77d797d2a4cfcb2e29c41ff6157bb6bba)?

This is due to a mistake I made: my `impl<T: TS, const N: usize> TS for [T; N]` just returns `Array<T>`...

escritorio-gustavo:
One question: Can `Vec::<T>::name` return `Array<T>`? Or is there a specific reason for it to return `Array`?

escritorio-gustavo:
Playing around with this further is causing conflicts with generic inlining, I'm getting `null`s instead of `T` and when trait bounds like `ToString` are involved I get `string` instead of `T`

escritorio-gustavo:
> Or is there a specific reason for it to return Array?

I think I got it, there's no way to resolve `T`'s possible generics, or if `name` or `inline` should be used

escritorio-gustavo:
> This is due to a mistake I made: my impl<T: TS, const N: usize> TS for [T; N] just returns Array<T>

I fixed this, but if I remove the changes to `macros/types/generics.rs` the issue with inline generics shows up

ts-rs/src/lib.rs
```diff
@@ -498,18 +498,60 @@ impl<T: TS> TS for Vec<T> {
...
+        [Dependency::from_ty::<T>()].into_iter().flatten().collect()
+    }
+
+    fn transparent() -> bool {
+        true
+    }
+}
+
+// Arrays longer than this limit will be emmited as Array<T>
+const ARRAY_TUPLE_LIMIT: usize = 100; // Temp value
```
escritorio-gustavo, line 518: Final value pending

escritorio-gustavo:
I managed to remove the conversion to a tuple and resolve the generic inlining problem, still results in a separate case for the array though

ts-rs/src/lib.rs
NyxCode: 100 sounds alright, though I always like powers of two :D

ts-rs/tests/generics.rs
```diff
@@ -278,5 +278,6 @@ fn trait_bounds() {
         t: [T; N],
     }
 
-    assert_eq!(D::<&str, 41>::decl(), "type D<T> = { t: Array<T>, }")
+    let ty = format!("type D<T> = {{ t: [{}], }}", "T, ".repeat(41).trim_end_matches(", "));
```
NyxCode, line 281: LOL 😅

NyxCode:
> > This is due to a mistake I made: my impl<T: TS, const N: usize> TS for [T; N] just returns Array
> 
> I fixed this, but if I remove the changes to `macros/types/generics.rs` the issue with inline generics shows up

This generics code always gives me a headache. Do you have an example of what exactly breaks with just the `impl TS for [T; N]` alone?

NyxCode:
> I didn't realize that was the case, I saw they were converted into a struct with unnamed fields and just assumed they worked for every size, since I couldn't find a limit in macros/src/types/tuple.rs.
Yeah, so I'm not aware of any way to implement a trait for all homogeneous tuples of one type. If there was a spread syntax for tuples or something like that, i think it'd be possible to do it inductively/recursively, but there isnt.

escritorio-gustavo:
> This generics code always gives me a headache. Do you have an example of what exactly breaks with just the impl TS for [T; N] alone?

I also get confused a lot, proc macros are really hard to keep up with, but I get the following failing test:
```
test free ... ok
test newtype ... FAILED
test interface ... FAILED

failures:

---- newtype stdout ----
thread 'newtype' panicked at ts-rs\tests\arrays.rs:24:5:
assertion `left == right` failed
  left: "[]"
 right: "[number, number, number, number]"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- interface stdout ----
thread 'interface' panicked at ts-rs\tests\arrays.rs:16:5:
assertion `left == right` failed
  left: "{ a: [], }"
 right: "{ a: [number, number, number, number], }"


failures:
    interface
    newtype
```

ts-rs/src/lib.rs
escritorio-gustavo: I can make it 128 lol

NyxCode:
Ah, okay, I think I (kinda) understand it now. Took me a bit to (again) understand how we hacked generics support into the lib ^^.
I don't like how it currently works, but any attempt at really fixing it would be super involved i'm afraid.  

That being said, this PR is strictly an improvement, so from my side, i'd be happy to merge this. Do you think it's ready, or is there something you'd still like to do?  

Great work, btw! Given the technical debt, I like where this ended up - doing most of the work in the `[T; N]` impl, and only pulling out the `T` and formatting it to get generics working. Can't think of any better way to do this given the current constraints.

NyxCode:
> It would be really helpful if the docs on proc macros were more in depth, this stuff is really hard to figure out at times
I dont disagree with that, though we're doing pretty weird stuff here.  
If I would re-write this library from scratch, I think I'd try to put more of the logic into the actual runtime, and less into the proc macros. Basically just the `TS` trait to provide an AST for the type, and then doing the actual TypeScript codegen during the test.  Not sure if that would work out, but if it did, I'm sure it'd be much nicer. 
This mixing of runtime and proc-macro-time we're doing ("This function returns a tokenstream that expands to an expression that returns a string ...") is just painfull to work with.

PR #211, opened 2024-01-26
Always convert to absolute path before calculating diff

### LINKED ISSUE #111: generating invalid TypeScript import path with using Cargo workspace
Hi, thanks for your excellent work. I love ts-rs.

I use ts-rs with the project which uses Carog workspace. I found the following issue.
If that project has crates on both the grandchildren directory and child directory, it may generate an invalid TypeScript import path.
I created [example project](https://github.com/h-michael/ts-rs-cargo-workspace-example).
[This struct](https://github.com/h-michael/ts-rs-cargo-workspace-example/blob/3b0b675606fc8e4cfc1e6d693a7f78310e7cb0a4/api/src/lib.rs#L1-L13) uses another [crate's struct](https://github.com/h-michael/ts-rs-cargo-workspace-example/blob/3b0b675606fc8e4cfc1e6d693a7f78310e7cb0a4/dep_a/lib_a/src/lib.rs#L3-L8) that crates on the grandchildren directory.
In this case, [invalid import paths](https://github.com/h-michael/ts-rs-cargo-workspace-example/blob/3b0b675606fc8e4cfc1e6d693a7f78310e7cb0a4/bindings/ABC.ts#L2-L3) are generated.
That should be as follows.

```typescript
import type { A } from "./A";
import type { B } from "./B";
```

I know this is a corner case. I can avoid this by using a flattened directory architecture. But actually, I use ts-rs with a bigger mono repository than the above example. So I want to use nested directories to categorize crates.

If you can give me some guidelines to fix this problem, I can create PR to fix it. Thanks.

escritorio-gustavo:
Convert `path` and `base` into absolute paths before diffing the two

Fixes #111

### DIFF
--- a/ts-rs/src/export.rs
+++ b/ts-rs/src/export.rs
@@ -144,39 +144,43 @@ where
     let path = path.as_ref();
     let base = base.as_ref();
 
-    if path.is_absolute() != base.is_absolute() {
-        if path.is_absolute() {
-            Some(PathBuf::from(path))
-        } else {
-            None
-        }
+    let path = if path.is_absolute() {
+        path.to_path_buf()
     } else {
-        let mut ita = path.components();
-        let mut itb = base.components();
-        let mut comps: Vec<Component> = vec![];
-        loop {
-            match (ita.next(), itb.next()) {
-                (None, None) => break,
-                (Some(a), None) => {
-                    comps.push(a);
-                    comps.extend(ita.by_ref());
-                    break;
-                }
-                (None, _) => comps.push(Component::ParentDir),
-                (Some(a), Some(b)) if comps.is_empty() && a == b => (),
-                (Some(a), Some(Component::CurDir)) => comps.push(a),
-                (Some(_), Some(Component::ParentDir)) => return None,
-                (Some(a), Some(_)) => {
+        std::env::current_dir().unwrap().join(path)
+    };
+
+    let base = if base.is_absolute() {
+        base.to_path_buf()
+    } else {
+        std::env::current_dir().unwrap().join(base)
+    };
+    
+    let mut ita = path.components();
+    let mut itb = base.components();
+    let mut comps: Vec<Component> = vec![];
+    loop {
+        match (ita.next(), itb.next()) {
+            (None, None) => break,
+            (Some(a), None) => {
+                comps.push(a);
+                comps.extend(ita.by_ref());
+                break;
+            }
+            (None, _) => comps.push(Component::ParentDir),
+            (Some(a), Some(b)) if comps.is_empty() && a == b => (),
+            (Some(a), Some(Component::CurDir)) => comps.push(a),
+            (Some(_), Some(Component::ParentDir)) => return None,
+            (Some(a), Some(_)) => {
+                comps.push(Component::ParentDir);
+                for _ in itb {
                     comps.push(Component::ParentDir);
-                    for _ in itb {
-                        comps.push(Component::ParentDir);
-                    }
-                    comps.push(a);
-                    comps.extend(ita.by_ref());
-                    break;
                 }
+                comps.push(a);
+                comps.extend(ita.by_ref());
+                break;
             }
         }
-        Some(comps.iter().map(|c| c.as_os_str()).collect())
     }
+    Some(comps.iter().map(|c| c.as_os_str()).collect())
 }


escritorio-gustavo:
I used [this StackOverflow answer](https://stackoverflow.com/a/54817755/22372249) to implement path conversion, but it seems to me that the current diffing logic already does the job of the `path-clean` crate, so I did not add it. Please correct me if I am mistaken about this

NyxCode:
Code-wise, this looks alright! I'd have to write a few test to really understand what currently goes wrong, and how this fixes it though. Maybe you could provide some info on this.

NyxCode:
Ah, awesome, missed that test repository! I think now I understand the issue!
Might be cool to be able to reproduct it in a unit test, right in `export.ts`.
My gut tells me that we have a `..` too much in some cases. Without that, `import type { A } from "../../bindings/A"` should turn into `import type { A } from "../bindings/A"`, which should turn into `import type { A } from "./A"`.

NyxCode:
Let me know if you get stuck, happy to take a look at it.

escritorio-gustavo:
I have found that the command `cargo locate-project --workspace` gives the path to the workspace's `Cargo.toml`, and when not in a workspace, the project's `Cargo.toml`. We could make relative paths relative to that, but that would be a breaking change

escritorio-gustavo:
In @h-michael's case, all of his exports would be `#[export_to = "./bindings/"]`

escritorio-gustavo:
Another issue with this approach: the command returns a json string (`{"root":"C:\\Users\\...\\ts-rs\\Cargo.toml"}`) , so we would have to either
- take a dependency on `serde_json` or
- `.trim_start_matches(r#"{"root":""#).trim_end_matches(r#""}"#)` which is **very** hacky

escritorio-gustavo:
Another issue with this approach: the command returns a json string (`{"root":"C:\\Users\\...\\ts-rs\\Cargo.toml"}`) , so we would have to either
- take a dependency on `serde_json` or
- `.trim_start_matches(r#"{"root":""#).trim_end_matches(r#""}"#)` which is **very** hacky

NyxCode:
> Another issue with this approach: the command returns a json string (`{"root":"C:\\Users\\...\\ts-rs\\Cargo.toml"}`) , so we would have to either
> 
> * take a dependency on `serde_json` or
> * `.trim_start_matches(r#"{"root":""#).trim_end_matches(r#""}"#)` which is **very** hacky

There's `--message-format plain`, which is what we want, i think.

I guess we should decide if we want to make everything relative to the workspace root. It's probably the better choice, but, as you mentioned, might be bothersome to the users. I'm not sure how big of a deal this would be, since the current state is somewhat broken anyway.

That being said, I think we could also fix the current implementation. I'm not 100% sure yet how, but:  
We create the import path [here](https://github.com/Aleph-Alpha/ts-rs/blob/a38ac4b996d3f1b2dff5772050dd0b33f4d45789/ts-rs/src/export.rs#L110).
Here, we might want to modify either `from` or `import` to reflect that these two paths are relative, but **not relative to the same directory.**  
We'd still need a way to figure out how to get the "crate path" of a given `Dependency`.

escritorio-gustavo:
> There's --message-format plain, which is what we want, i think.

Just checked, you're right, that gives the path in plaintext

> We'd still need a way to figure out how to get the "crate path" of a given Dependency.

This seems **really** hard to do. When `cargo test` gets to the `api` project, it runs whatever it needs from `dep_a` and `dep_b`, using its own `Cargo.toml` as the root, which gets to `Dependency` and eventually breaks the `import` path. I am not sure there is a way to get a diferent crate's root path to allow this to work

escritorio-gustavo:
> There's `--message-format plain`, which is what we want, i think.

We should also add `--quiet` to avoid warnings and notes about resolver versions

NyxCode:
> > We'd still need a way to figure out how to get the "crate path" of a given `Dependency`.
> 
> This seems **really** hard to do. When `cargo test` gets to the `api` project, it runs whatever it needs from `dep_a` and `dep_b`, using its own `Cargo.toml` as the root, which gets to `Dependency` and eventually breaks the `import` path. I am not sure there is a way to get a diferent crate's root path to allow this to work

I might not have thought this through 100%, but wouldn't this work?
- The `#[derive(TS)]` macro get's executed for each type within each crate. 
- In that macro invocation, we've got access to `CARGO_MANIFEST_DIR`, which should give us the path to the crate
- When we generate the `impl TS for T { .. }` impl, we add a `const CRATE_PATH: &'static str`.
- We add a `crate_path: &'static str` to `Dependency` as well, and during `Dependency::from_ty<T>()`, we set it to `T::CRATE_PATH`

> We should also add --quiet to avoid warnings and notes about resolver versions

Probably a good idea, though that might end up in stderr, idk

escritorio-gustavo:
As far as I've been able to test/understand (which, to be honest, is not very far), this doesn't work because when `api` goes to get the import path, all the code executes taking `api` as the root.
The `const` is also problematic because you can't assign a `const` to something that reads the file system

Again, I haven't completely wrapped my head around this quite yet, so I may very well be wrong, but it looks like this is what's happening

escritorio-gustavo:
> In that macro invocation, we've got access to CARGO_MANIFEST_DIR, which should give us the path to the crate

I haven't tried anything using this. That might actually work

escritorio-gustavo:
I have tried today and it doesn't work. The problem is, when calculating the import path, CARGO_MANIFEST_DIR evaluates to the `api` directory, regardless of which crate the struct is from

escritorio-gustavo:
> The const is also problematic because you can't assign a const to something that reads the file system

I tried to use `std::env!` to get around that and that gives me the `ts-rs/ts-rs` directory

escritorio-gustavo:
I tried this today, and after a few failed attempts with commically incorrect results I actually got it to work, but it does require cleaning up the paths with something like `path-clean`

escritorio-gustavo:
> but it does require cleaning up the paths with something like path-clean

Nevermind, `Path::canonicalize` seems to work just fine

escritorio-gustavo:
@NyxCode, I think we have four options here
- Use `canonicalize` anyway
- Copy `path-clean`'s implementation
- Take a dependency on `path-clean`
- Somehow change diffing logic to deal with a bunch of `../` components

NyxCode:
@escritorio-gustavo Awesome, thank you so much for the research on this.
To get a full picture of this issue (and the related #177), I think I'll setup some end-to-end tests for the different scenarios.  

Let's say my crate `MyRestAPI` is depending on some other crate which implements `TS` on some of its types. If `MyRestAPI` has structs which include this type, the dependencies are broken. This is what I understand #177 to be about.

I'd really like a solution which works for both workspaces and the scenario I just described.

escritorio-gustavo:
Oh sorry, I missed this bit of #177
> "CoolEnum" also has the correct annotations, but is being imported from another crate

Which made me think it was about the `#[ts(as = "...")]` attribute

escritorio-gustavo:
> I think I'll setup some end-to-end tests for the different scenarios.

Thanks, due to the fact that this whole problem relies on having multiple crates involved I'm really struggling to come up with a way to test it

escritorio-gustavo:
> Which made me think it was about the #[ts(as = "...")] attribute

Still... can't `as` solve the external crate case?

escritorio-gustavo:
> Still... can't as solve the external crate case?

Something like this:
```rs
#[derive(TS)]
#[ts(export)]
pub struct LibraryTypeDef {
    pub a: i32
}

pub struct MyStruct {
    #[ts(as = "LibraryTypeDef")]
    pub foo: other_crate::LibraryType
}
```

NyxCode:
Absolutely, that'd work as a workaround. But If `other_crate::LibraryType` already implements `TS`, it'd be nice to be able to just use that.

escritorio-gustavo:
It really would be nice, but should library code ever implement `TS`? Seems like this should be an application code decision (especially the extra dependency + tests)

Issue #112, opened 2022-08-11
Optional Options should have a `null` type


ryanw:
Currently this:

```rust
struct Foo {
  bar: Option<String>
}
```

Will generate:

```typescript
interface Foo {
  bar: string | null,
}
```

But this:

```rust
struct Foo {
  #[ts(optional)]
  bar: Option<String>
}
```

Will generate:
```typescript
interface Foo {
  bar?: string,
}
```

This means it's not possible to assign the required value to the `#[ts(optional)]` one as you end up with the error: 

```
Type 'null' is not assignable to type 'string | undefined'
```

It seems to be me that the `#[ts(optional)]` attribute shouldn't change the type at all, and simply adds the `?` to the field. i.e. It should spit out:

```typescript
interface Foo {
  bar?: string | null,
}
```

petersn:
I agree that `| null` and `?` are sort of orthogonal, and probably there should be finer grained control here. Analogously, currently `#[serde(skip_serializing_if = "...")]` only checks for `Option::is_none`, because ts-rs's "optional" field simply means replacing an `Option<T>` field with a `T` field and slapping a `?` on it. This means that ts-rs doesn't emit `?`s for other `skip_serializing_if` fields, because it can't know that they're on Option types.

I propose we replace this with two concepts: optional (`?`), and nullable (`| null`). By default `Option<T>` produces a nullable field, and `skip_serializing_if` always produces an optional field. We can even make it so that if it detects `skip_serializing_if = "Option::is_none"` that it becomes optional and _not_ nullable, which would make this proposal be a backwards-compatible extension of the current behavior. 

I have a patch that does something sort of like this (https://github.com/Aleph-Alpha/ts-rs/compare/main...petersn:ts-rs:optional-changes although it's mixed in with other changes), but I'm curious what the maintainers think is the right behavior here.

NyxCode:
I started writing a reply, and while writing it realized how complex the different tradeoffs are here. 
After thinking about this, I think I've come to a conclusion:

# Interactions with `serde-compat`
- serde attributes should get "translated" into `#[ts(..)]` attributes.  
- If, given a set of serde attributes, there are multiple possible types we could generate, I came to the conclusion that we should chose the *smallest* type (*small* in a set-theoretical way).  
- Finally, that default must be able to be overwritten.

The serde attributes I'm thinking of here are
- `#[serde(default)]`
- `#[serde(skip_serializing_if = "Option::is_none")]`

[...]

PR #213, opened 2024-01-29
Implement `#[ts(optional = nullable)]` 

NyxCode:
This PR implements `#[ts(optional = nullable)]`, as discussed in #112.
`#[ts(optional)] Option<T>` still generates `t?: T`, while `#[ts(optional = nullable)]` generates `t?: T | null`.

I plan on following this up at some point with support for the serde attributes outlined in #112.

### DIFF
--- a/macros/src/attr/field.rs
+++ b/macros/src/attr/field.rs
@@ -1,7 +1,9 @@
 use syn::{Attribute, Ident, Result};
+use syn::spanned::Spanned;
+
+use crate::utils::parse_attrs;
 
 use super::parse_assign_str;
-use crate::utils::parse_attrs;
 
 #[derive(Default)]
 pub struct FieldAttr {
@@ -9,8 +11,17 @@ pub struct FieldAttr {
     pub rename: Option<String>,
     pub inline: bool,
     pub skip: bool,
-    pub optional: bool,
+    pub optional: Optional,
     pub flatten: bool,
+}
+
+/// Indicates whether the field is marked with `#[ts(optional)]`.
+/// `#[ts(optional)]` turns an `t: Option<T>` into `t?: T`, while
+/// `#[ts(optional = nullable)]` turns it into `t?: T | null`.
+#[derive(Default)]
+pub struct Optional {
+    pub optional: bool,
+    pub nullable: bool,
 }
 
 #[cfg(feature = "serde-compat")]
@@ -36,7 +47,7 @@ impl FieldAttr {
             rename,
             inline,
             skip,
-            optional,
+            optional: Optional { optional, nullable },
             flatten,
         }: FieldAttr,
     ) {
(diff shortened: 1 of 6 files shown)


NyxCode:
Closes #112

NyxCode:
@escritorio-gustavo would love to hear what you think about this!

escritorio-gustavo:
This is awesome! I thought there would be a separate `#[ts(nullable)]` attribute, but making it an argument for `#[ts(optional)]` is way better to prevent it from being used instead of an actual `Option`

escritorio-gustavo (approved):

escritorio-gustavo:
Quick question, is the `optional` field needed? It always true unless the attribute is not present. This could an `Option` with just one field for `nullable`
```rs
#[derive(default)]
pub struct Nullable(pub bool);

#[derive(Default)]
pub struct FieldAttr {
    pub type_override: Option<String>,
    pub type_override: Option<String>,
    pub rename: Option<String>,
    pub rename: Option<String>,
    pub inline: bool,
    pub inline: bool,
    pub skip: bool,
    pub skip: bool,
    pub optional: bool,
    pub optional: Option<Nullable>,
    pub flatten: bool,
    pub flatten: bool,
}

let (ty, optional_annotation) = match optional {
      Some(Nullable(nullable) => {
          let inner_type = extract_option_argument(&field.ty)?; // inner type of the optional
          match nullable {
              true => (&field.ty, "?"),  // if it's nullable, we keep the original type
              false => (inner_type, "?"), // if not, we use the Option's inner type
          }
      },
      None => (&field.ty, "")
 };
```

This is just an idea though, I'm totally fine with merging this as is

NyxCode:
> Quick question, is the `optional` field needed? It always `true` unless the attribute is not present. This could an `Option` with just one field for `nullable`
> 
> ```rust
> #[derive(default)]
> pub struct Nullable(pub bool);
> 
> #[derive(Default)]
> pub struct FieldAttr {
>     pub type_override: Option<String>,
>     pub type_override: Option<String>,
>     pub rename: Option<String>,
>     pub rename: Option<String>,
>     pub inline: bool,
>     pub inline: bool,
>     pub skip: bool,
>     pub skip: bool,
>     pub optional: bool,
>     pub optional: Option<Nullable>,
>     pub flatten: bool,
>     pub flatten: bool,
> }
> 
> let (ty, optional_annotation) = match optional {
>       Some(Nullable(nullable) => {
>           let inner_type = extract_option_argument(&field.ty)?; // inner type of the optional
>           match nullable {
>               true => (&field.ty, "?"),  // if it's nullable, we keep the original type
>               false => (inner_type, "?"), // if not, we use the Option's inner type
>           }
>       },
>       None => (&field.ty, "")
>  };
> ```
> 
> This is just an idea though, I'm totally fine with merging this as is

Sure, that'd work as well. I tried a couple of options, and the one i ended up with felt the least ugly to me.  
If you've got strong opinions on this though, feel free to change it.

escritorio-gustavo:
> Sure, that'd work as well. I tried a couple of options, and the one i ended up with felt the least ugly to me.
> If you've got strong opinions on this though, feel free to change it.

I don't think there's a need for that, I was just curious

PR #218, opened 2024-01-31
[WIP] E2E tests

NyxCode:
This PR adds the directory `e2e`, which is supposed to include end-to-end tests to make sure imports / exports are working.

I added these two tests to illustrate how the current default behaviour is somewhat broken:

### [dependencies](./dependencies)
A user creates the crate `consumer`, which depends on some crate `dependency1`, which possibly lives on [crates.io](https://crates.io). 
If a struct in `consumer` contains a type from `dependency1`, it should be exported as well.

### [workspace](./workspace)
A user creates a workspace, containing `crate1`, `crate2`, and `parent`.  
`crate1` and `crate2` are independent, but `parent` depends on both `crate1` and `crate2`.

CI runs `cargo test` and `tsc bindings/*` to check if the output is at least valid, which it currently **is not** for both tests.


e2e/dependencies/dependency1/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+use ts_rs::TS;
+
+#[derive(TS)]
+pub struct LibraryType {
+    pub a: i32
+}
```
escritorio-gustavo, line 6: Interesting, since this type doesn't contain `#[export]` we'd have to come up with some other way to export it in the consumer crate

e2e/workspace/crate2/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+use ts_rs::TS;
+
+#[derive(TS)]
+pub struct Crate2 {
+    pub x: i32
+}
```
escritorio-gustavo, line 6: I believe I've got this to work when `#[export, export_to = "..."]` is used

e2e/dependencies/dependency1/src/lib.rs
NyxCode: Right! I'm not sure about how to exactly solve this yet - maybe a `#[ts(export)]` type should cause all of its dependencies to be exported as well?

e2e/workspace/crate2/src/lib.rs
NyxCode: Yeah, if all crates of the workspace use `#[ts(export_to = "../some_dir")]`, then they all put their files in one directory & the imports work.

e2e/dependencies/dependency1/src/lib.rs
escritorio-gustavo: We'd have to make sure it's only exported once, even if the same type is used in a bunch of places

escritorio-gustavo: But this might be confusing, as the user is not explicitly `#[ts(export)]`ing the type

e2e/workspace/crate2/src/lib.rs
escritorio-gustavo: Correct, and since they are all in the same workspace, it should be easy enough to ensure they are in the same directory.
Although, due to the changes to path diffing, I don't think they even **need** to be... just `#[ts(export)]` might be enough

e2e/dependencies/dependency1/src/lib.rs
NyxCode: > But this might be confusing, as the user is not explicitly `#[ts(export)]`ing the type

The problem in that scenario is that the user just *cant* explicitly export it, since it's in an other crate that may not be under his control.

> We'd have to make sure it's only exported once, even if the same type is used in a bunch of places

That's true, this might be the biggest challenge here.

escritorio-gustavo: > The problem in that scenario is that the user just cant explicitly export it, since it's in an other crate that may not be under his control.

That is true, and it brings up the question I asked in #211:

> Should library code in `crates.io` ever implement TS? Seems like this should be an application code decision (especially the extra dependency + tests)

e2e/workspace/crate2/src/lib.rs
escritorio-gustavo: > just #[ts(export)] might be enough

Haven't tested this yet though

e2e/dependencies/dependency1/src/lib.rs
NyxCode: (let's maybe have this discussion here then)
> Should library code in crates.io ever implement TS? Seems like this should be an application code decision (especially the extra dependency + tests)

I think it'd be a good idea to support that, yeah! These libraries would probably never `#[ts(export)]`, but making their types implement `TS` might be usefull in some cases, I think.

escritorio-gustavo: > That is true, and it brings up the question I asked in https://github.com/Aleph-Alpha/ts-rs/pull/211

Speaking of which, this is getting hard to follow across two PRs, I opened a discussion (#219) to centralize this

e2e/workspace/crate2/src/lib.rs
escritorio-gustavo: > Haven't tested this yet though

Tested it now, it works!

PR #220, opened 2024-02-01
e2e + #195 test

escritorio-gustavo:
@NyxCode I have copied over the code from #195 to check it if'd be able to pass the e2e tests. With a change to the `dependencies` test (adding `cargo t -p dependency1`) it works!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,6 +12,7 @@ jobs:
       - name: dependencies e2e test
         working-directory: e2e/dependencies/consumer
         run: |
+          cargo t -p dependency1
           cargo t 
           tsc bindings/* --noEmit
   e2e-workspace:
--- a/e2e/dependencies/dependency1/src/lib.rs
+++ b/e2e/dependencies/dependency1/src/lib.rs
@@ -1,6 +1,12 @@
 use ts_rs::TS;
 
 #[derive(TS)]
+#[ts(export)]
 pub struct LibraryType {
     pub a: i32
+}
+
+#[test]
+fn env_set() {
+    assert_ne!(env!("TS_RS_EXPORT_DIR"), "");
 }
--- a/e2e/workspace/.cargo/config.toml
+++ b/e2e/workspace/.cargo/config.toml
@@ -1,0 +1,2 @@
+[env]
+TS_RS_EXPORT_DIR = { value = "./bindings/", relative = true }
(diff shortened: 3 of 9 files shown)


escritorio-gustavo:
Note: **all** changes outside `e2e` are from #195

e2e/dependencies/dependency1/src/lib.rs
```diff
@@ -1,6 +1,12 @@
 use ts_rs::TS;
 
 #[derive(TS)]
+#[ts(export)]
```
escritorio-gustavo, line 4: Lib authors must use `#[ts(export)]`

e2e/dependencies/dependency1/src/lib.rs
```diff
@@ -1,6 +1,12 @@
...
 pub struct LibraryType {
     pub a: i32
-}
\ No newline at end of file
+}
+
+#[test]
+fn env_set() {
+    assert_ne!(env!("TS_RS_EXPORT_DIR"), "");
+}
```
escritorio-gustavo, line 12: It is possible to enforce that the consumer sets the environment variable

.github/workflows/test.yml
```diff
@@ -12,6 +12,7 @@ jobs:
       - name: dependencies e2e test
         working-directory: e2e/dependencies/consumer
         run: |
+          cargo t -p dependency1
```
escritorio-gustavo, line 15: Need to run the dependencies' tests, luckily this only needs to be done on dependency install/update

e2e/workspace/.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[env]
+TS_RS_EXPORT_DIR = { value = "./parent/bindings/", relative = true }
```
escritorio-gustavo, line 2: The path is relative to `.cargo`'s parent

escritorio-gustavo (commented):
How it works:

escritorio-gustavo:
Btw, this makes #211 obsolete, so if this gets merged that PR should be closed

NyxCode:
Awesome!
(Please correct me if I get any of this wrong)
To me, it seems like these are the tradeoffs for this solution:
- **CON** library types must be `#[ts(export)]`.
- **CON** All library types will be exported, not just those used in the consumer
- **PRO** No locking is required, since every type is only ever exported by one test
- **---** Setting the env variables and running `cargo t -p` for every dependency is a bit cumbersome, so this solution would profit from a CLI tool that automatically does all that

escritorio-gustavo:
Yes, those are all correct

escritorio-gustavo:
> - **CON** All library types will be exported, not just those used in the consumer

If we create a CLI to run the tests, maybe it could filter only the types used in the consumer?

PR #225, opened 2024-02-05
Add support for `#[serde(rename_all_fields = "...")]` attribute

escritorio-gustavo:
Adds support for the `#[serde(rename_all_fields = "...")]` enum attribute, as well as a `#[ts(rename_all_fields = "...")]` equivalent.
This attribute behaves the same as adding `#[ts(rename_all = "...")]` to each **struct variant** of an enum, **but not on the enum itself**, so the variant names will not be affected, for that `#[ts(rename_all = "...")]` still must be used on the enum

### DIFF
--- a/macros/src/types/enum.rs
+++ b/macros/src/types/enum.rs
@@ -68,6 +68,10 @@ fn format_variant(
     generics: &Generics,
 ) -> syn::Result<()> {
     let variant_attr = VariantAttr::from_attrs(&variant.attrs)?;
+    let variant_attr = VariantAttr {
+        rename_all: variant_attr.rename_all.or(enum_attr.rename_all_fields),
+        ..variant_attr
+    };
 
     if variant_attr.skip {
         return Ok(());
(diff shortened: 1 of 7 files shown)


macros/src/types/enum.rs
```diff
@@ -68,6 +68,10 @@ fn format_variant(
     generics: &Generics,
 ) -> syn::Result<()> {
     let variant_attr = VariantAttr::from_attrs(&variant.attrs)?;
+    let variant_attr = VariantAttr {
+        rename_all: variant_attr.rename_all.or(enum_attr.rename_all_fields),
+        ..variant_attr
+    };
```
escritorio-gustavo, line 74: I shadowed the variable to avoid adding a `mut`, do you think this is a good approach?

NyxCode: Seems reasonable to me!
However, I think I'd prefer changing the `VariantAttr` constructor instead. Maybe renaming it to `VariantAttr::new`, and having it take `&[Attribute]` as well as `&EnumAttr`. I'm open to alternatives, but I kinda dislike that it takes two steps to create a `VariantAttr` now.

NyxCode (changes requested):
Besides that small nit-pick, nice and elegant change! Great work!

macros/src/types/enum.rs
escritorio-gustavo: Done :D

PR #226, opened 2024-02-05
Add support for `#[serde(untagged)]` on enum variants

escritorio-gustavo:
As of https://github.com/serde-rs/serde/pull/2403 serde supports adding `#[serde(untagged)]` to individual enum variants, this PR adds support for it to `ts-rs`


escritorio-gustavo:
@NyxCode, writing this made me wonder... should we implement `#[ts(tag = "...")]`, `#[ts(tag = "...", content = "...")]` and `#[ts(untagged)]`?

NyxCode (approved):
Good work, thanks!  
I think enum.rs:136 is not formatted correctly.

Other than that, nice & elegant change!

PR #228, opened 2024-02-09
feat: Add functionality to export doc strings on struct fields and enum variants

timephy:
This is an "extension" of #187 and tries to also export doc strings not only on "full types" but also on the fields/variants of structs/enums.

## Example

```rust
#[derive(TS)]
struct A {
    /// Doc of field
    ///
    /// Testing
    name: String,
}
```

is exported as

```ts
export interface A {
    /**
     * Doc of field
     *
     * Testing
     */
    name: string,
}
```

### JSDocs are only shows when starting in a new line

image
image

This of course interferes with the "format" feature and because of the requirement to start in a new line, all tests I guess.

### JSDocs are not applicable to unions (enums)

See https://github.com/microsoft/tsdoc/issues/164

This **might** depend on whether the enum is tagged internally/externally, etc.


escritorio-gustavo:
Hey @timephy! It seems this PR contains duplicates of the commits in #187. I think it would be a good idea to sync your main branch to this repo's and then create a new branch to avoid this issue

escritorio-gustavo:
## Notes:

- Generated TS now creates `type`s instead of interfaces
> ```ts
> export interface A {
>     /**
>      * Doc of field
>      *
>      * Testing
>      */
>     name: string,
> }
> ```

Should be 
> ```ts
> export type A = {
>     /**
>      * Doc of field
>      *
>      * Testing
>      */
>     name: string,
> }
> ```


> ### JSDocs are only shows when starting in a new line
> This of course interferes with the "format" feature and because of the requirement to start in a new line, all tests I guess.

Great observation! This likely means that in case the format feature is off we will have to avoid generating field docs, as they will not be rendered

> ### JSDocs are not applicable to unions (enums)
> This **might** depend on whether the enum is tagged internally/externally, etc.

Sadly there is no reasonable way to document enum variants, regardless of tagging, we should keep this to struct fields (which, on the bright side, will affect fields inside an enum's struct variant)

PR #229, opened 2024-02-10
feat: Add functionality to export doc strings on struct fields and enum variants

timephy:
This is an "extension" of #187 and tries to also export doc strings not only on "full types" but also on the fields/variants of structs/enums.

This is also a new PR as a replacement for #228, because that PR had a "wrong commit base".

## Example

```rust
#[derive(TS)]
struct A {
    /// Doc of field
    ///
    /// Testing
    name: String,
}
```

is exported as (with `format` feature)

```ts
export type A = {
  /**
   * Doc of field
   *
   * Testing
   */
  name: string;
};
```

and 

```rust
enum F {
    /// Doc of variant
    ///
    /// Testing
    VarA,
    /// Doc of variant
    ///
    /// Testing
    VarB(),
    /// Doc of variant
    ///
    /// Testing
    VarC {
        /// Doc of field of variant
        ///
        /// Testing
        variant_vield: i32,
    },
}
```

is exported as (with `format` feature)

```ts
export type F = "VarA" | { "VarB": never[] } | {
  "VarC": {
    /**
     * Doc of field of variant
     *
     * Testing
     */
    variant_vield: number;
  };
};
```

## Notes

These are the "problems" that have to be thought of.

Following notes are a combination of my own observations + input from @escritorio-gustavo.

### JSDocs are only shown when starting in a new line (solved)

image
image

Inserted doc strings always have a leading newline to solve this.

### JSDocs are not applicable to unions (enums)

See https://github.com/microsoft/tsdoc/issues/164

It seems like this is currently not possible, not even depending on the tagging-kind of enums.


timephy:
I would like to ask for your help @escritorio-gustavo to solve these issues 🙏

- I am unsure about how or even if we should handle docs for newtypes and tuples...
- I have copied the function `format_docs` because I needed it inside the `macros` module, but it is defined inside the `ts-rs` module, and therefore not importable..

Should we create a new `utils` crate?

I believe there is no way to refactor the usage of `format_docs` from inside `ts-rs` to inside `macros` for the doc generation for "full types", because the doc string needs to be in front of typescript's `"export"`, which is generated from inside `ts-ts`... But maybe you can find a way - although we can of course just keep it this way.

image

timephy:
(the commit message "exporting docs on struct fields not works" should of course be "*now* works", lol)

escritorio-gustavo:
> I am unsure about how or even if we should handle docs for newtypes and tuples...

Good question, I assume they currently don't export the docs right? I think we can skip documenting fields for these types as documenting the field of a newtype is kinda pointless and I'm not sure if TSDoc even renders tuple member documentation

> You may also please have a look how this implementation behaves when using flatten, which I did not look into yet

Sure, I'll write some tests and push them to the PR

> I have copied the function format_docs because I needed it inside the macros module, but it is defined inside the ts-rs module, and therefore not importable..
> Should we create a new utils crate?

That could work, but we need to check with @NyxCode.

It's really annoying that we can't just move the function to `macros` and export it. Maybe we could turn format_docs into a proc_macro to do this? Seems like overkill but could also be an option

macros/src/attr/variant.rs
```diff
@@ -13,6 +13,8 @@ pub struct VariantAttr {
     pub inline: bool,
     pub skip: bool,
     pub untagged: bool,
+    // `docs` is parsed, but JSDocs don't work on unions, therefore it's value is not used
+    pub docs: Vec<String>,
```
escritorio-gustavo, line 17: We should probably just remove this field. If ever support is added, we can come back to it

macros/src/types/newtype.rs
```diff
@@ -29,6 +29,7 @@ pub(crate) fn newtype(
         skip,
         optional,
         flatten,
+        docs: _, // TODO: Not sure what to do here with docs yet
```
escritorio-gustavo, line 32: I actually think discarding is the right thing to do, documenting the field of a newtype seems pointless to me, just document the type itself

escritorio-gustavo:
I think we can make `parse_docs` return the properly formatted docs as a `Result<String>` and eliminate `format_docs` entirely

escritorio-gustavo:
I think this is now ready for merging. Thank you so much for your work on this and #187 @timephy!

timephy:
Wonderful! Thanks for all the help! 🤝🏽

PR #231, opened 2024-02-10
Fix `#[ts(skip)]` in newtype variants

NyxCode:


macros/src/types/enum.rs
```diff
@@ -115,18 +115,20 @@ fn format_variant(
...
+                        (Some(_), Some(_)) => syn_err!("`type` is not compatible with `as`"),
+                        (Some(type_override), None) => quote! { #type_override },
+                        (None, Some(type_as)) => {
+                            format_type(&syn::parse_str::<Type>(&type_as)?, dependencies, generics)
+                        }
+                        (None, None) => format_type(&unnamed.unnamed[0].ty, dependencies, generics),
+                    };
+
                     quote!(format!("{{ \"{}\": \"{}\", \"{}\": {} }}", #tag, #name, #content, #ty))
                 }
```
NyxCode, line 133: @timephy This here was the issue: `format_type` was called outside of the `if skip { .. }`, adding the content to the dependencies even if the field was to be skipped.

timephy: Hey :)
Great investigation! Thank you for the quick fix - highly appreciated!

Issue #214, opened 2024-01-29
Generics don't work with `inline` and `flatten` when they have a default type or a trait bound


escritorio-gustavo:
As of #212, we can now properly inline a generic type, as long as the generic does not have trait bounds. We have added an ignored test to help future implementations.

[Trait bound](https://github.com/Aleph-Alpha/ts-rs/blob/1dc210bd2e531c77d2f06dd8386f4d952e7fb3ba/ts-rs/tests/generics.rs#L172)

## Why did #212 not implement this?
#212 bases its decision of inlining the generic type based on wether or not the provided type returns `"null"` from its `.inline()` method. That basically means we are checking for the unit type `()`.

The problem is this doesn't work for trait bounds, since if you have any given trait, like `ToString` it is unlikely that the unit type implements said trait, and since tuples are always foreign you cannot implement it yourself, since neither `ToString` nor `()` are defined by your crate.

## Solution
Somehow implement the trait bounds to the unit type? Honestly I have no idea, when I wrote this issue I had the crate's unit tests in mind, but now, from the perspective of `generate_export_test`, all that matters is the unit type
Maybe we can accept some `default_type` attribute that is a path to a struct defined in the user's crate that behaves the same as the unit type (`inline`s to `"null"`) and implements all necessary traits?

escritorio-gustavo:
Minor update, I was looking through the test generation code and the type used is always `()`, which means that this should work even with trait bounds if the following conditions are met:
- The user implements all relevant traits for `()`
- The check for trait bounds is removed

escritorio-gustavo:
Still no idea how to deal with default types though

escritorio-gustavo:
> - The user implements all relevant traits for `()` if they plan to use `#[export]`

And of course... the orphan rule makes this pretty much impossible

escritorio-gustavo:
I think I got it the wrong way around. When the type has a default but no trait bounds it should be fine.

The `#[ts(export)]` attribute overrides the default with `()`, so the export properly uses `T`, but when inlining whatever type is applied, even the default, should be used instead of `T`

NyxCode:
I'd love to find a solution to this. I think I get what's happening, but i'll have to really go through it all in detail to understand all the edge cases.  
[...]

Issue #232, opened 2024-02-12
'internal error: entered unreachable code' when generating struct fields using Results


Tom-Goring:
When doing something like:

```rust
use ts_rs::TS;

#[derive(TS)]
#[ts(export)]
struct State {
    a: Vec<Result<String, String>>,
    b: Vec<Result<String, String>>,
}
```

I get the error 'internal error: entered unreachable code' - from `lib.rs:520:9` on main. (I'm using the latest commit).

This could be user error as if I do:

```rust
#[derive(TS)]
#[ts(export)]
struct State {
    #[ts(inline)]
    a: Vec<Result<String, String>>,
    #[ts(inline)]
    b: Vec<Result<String, String>>,
}
```

it works fine.

If this behaviour is intended it might be best to replace the use of `unreachable!()` for clarity?

On another note - using inline like this means no imports are generated for types that appear in the inline, e.g for:

```rust
use ts_rs::TS;

#[derive(TS)]
#[ts(export)]
struct State {
    #[ts(inline)]
    a: Vec<Result<EnumWithName, String>>,
    #[ts(inline)]
    b: Vec<Result<EnumWithName, String>>,
}

#[derive(TS)]
#[ts(export)]
struct EnumWithName {
    name: String,
    inner: Enum
}

#[derive(TS)]
#[ts(export)]
enum Enum {
    A,
    B
}
```

the generated type I get for `State` is:

```typescript
// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.

export type State = { a: Array<{ Ok : { name: string, inner: Enum, } } | { Err : string }>, b: Array<{ Ok : { name: string, inner: Enum, } } | { Err : string }>, }
```

The other files generate correctly (`Enum` and `EnumWithName`), but `Enum` is not imported and so the file isn't totally valid to compile. Is there a way around this, or is this behaviour or my use of `inline` here not intended.

Thanks for the crate as well - I've got a lot of use out of it!

NyxCode:
Thanks for opening the issue!  
I'm suprised our test suite didn't catch this, especially since this seems like a somewhat deeper issue.  
In the meantime, I don't see any other workaround besides using `#[ts(type = "Array<{ Ok: string } | { Err: string }>")]`.

escritorio-gustavo:
[...]

Issue #168, opened 2023-09-01
Import of type used in inlined generic type generated


jiftoo:
Hi. I noticed a bug or a limitation of this library. Consider this example:

```rust
use std::collections::HashMap;
use ts_rs::TS;

#[derive(TS)]
#[ts(export)]
pub struct Foo {
	#[ts(inline)]
	map: HashMap<usize, Bar>,
}

#[derive(TS)]
struct Bar {
	#[ts(inline)]
	map: HashMap<usize, Baz>,
}

#[derive(TS)]
struct Baz {
	#[ts(inline)]
	map: HashMap<usize, String>,
}
```

The .ts file produced by this snippet contains an `import type` statement referring to the first flattened generic type (`Bar` in this case), which is not used anywhere. No other imports are generated. Tested with `Vec` and `HashMap`.
```typescript
import type { Bar } from "./Bar";

export interface Foo { map: Record<number, { map: Record<number, { map: Record<number, string>, }>, }>, }
```

The expected output should not include the first line. Please correct me if I'm missing something from the docs.

PR #233, opened 2024-02-12
Bug: Flattening generic type also flattens all generic parameters

NyxCode:
Hey @escritorio-gustavo!
While trying to understand #232, I encountered this issue here. I'm not yet sure if it's the underlying cause, but maybe we should start with this.

Flattening an generic enum currently seems to not only flatten the enum, but also everything inside.  
I've added a test to highlight this behaviour.

Any ideas what's going on here?


escritorio-gustavo:
https://github.com/Aleph-Alpha/ts-rs/blob/ed9abd306720107702872264184f7649bc814aa4/macros/src/types/generics.rs#L66-L77

Seems to be the issue. Changing `inline` to `name` fixes the new test, but causes the `inline` test in the same file to fail

escritorio-gustavo:
Every other test seems to be fine though

NyxCode:
Ah, I see, thanks!
I managed to hack smth together that works, but I get the feeling that our current test suite cannot be fully trusted.  
I'll look into getting some fuzzing working, so that we can get a complete picture of what currently works and what doesn't.

escritorio-gustavo:
This is a very annoying bug. It's enterely caused by what I did in #215 because `Vec<T>` would export properly

gustavo-shigueo:
Hey @NyxCode, I'm at home right now, so I don't have access to my work account (@escritorio-gustavo).
I was trying something out and this seems to pass all the tests
```rs
return quote!(
    match <#generic_ident>::name().as_str() {
        "null" => #generic_ident_str.to_owned(),

        x if !std::any::type_name::<#generic_ident>().contains('<') => {
            x.to_owned()
        }

        _ => <#generic_ident>::inline()
    }
);
```

I am not sure if it is a good idea or if it breaks other stuff though

escritorio-gustavo:
> Hey @NyxCode, I'm at home right now, so I don't have access to my work account (@escritorio-gustavo).


Just confirming, this was actually me

escritorio-gustavo:
If the macro could somehow read the value of `type_name`, convert the string into the type path and call `name_with_type_args` on it, that'd be somewhat better, but sadly, I don't think this is possible

escritorio-gustavo:
> ```rust
> return quote!(
>     match <#generic_ident>::name().as_str() {
>         "null" => #generic_ident_str.to_owned(),
> 
>         x if !std::any::type_name::<#generic_ident>().contains('<') => {
>             x.to_owned()
>         }
> 
>         _ => <#generic_ident>::inline()
>     }
> );
> ```
> 
> I am not sure if it is a good idea or if it breaks other stuff though

This somewhat improves the situation, but still generates more inlining than expected. That is:
```rs
  #[derive(TS)]
  struct OtherType<T>(T);

  #[derive(TS)]
  struct SomeType<T>(OtherType<T>);

  #[derive(TS)]
  enum MyEnum<A, B> {
      VariantA(A),
      VariantB(B)
  }

  #[derive(TS)]
  struct Parent {
      e: MyEnum<i32, i32>,
      #[ts(inline)]
      e1: MyEnum<i32, SomeType<String>>
  }

  // This fails!
  // The #[ts(inline)] seems to inline recursively, so not only the definition of `MyEnum`, but
  // also the definition of `SomeType`.
  assert_eq!(
      Parent::decl(),
      "type Parent = { \
          e: MyEnum<number, number>, \
          e1: { \"VariantA\": number } | { \"VariantB\": SomeType<string> }, \
      }"
  );
```

Will fail, giving `OtherType<string>` instead of SomeType<string>

escritorio-gustavo:
I **think** it is no longer recursively inlining, but instead, it inlines one level deeper than it should. I'm really not sure though

NyxCode:
```ts
import type { LibraryType2<LibraryType2<ConsumerType>> } from "./LibraryType2";
import type { LibraryType2<LibraryType2<LibraryType1>> } from "./LibraryType2";
```
whoops 😆

NyxCode:
Not only are the import names broken, but our original logic of de-duping them doesn't work anymore (did it ever work with generics?).
https://github.com/Aleph-Alpha/ts-rs/blob/5daeed248dc33ad3a88f94a4deab107c91ba34db/ts-rs/src/export.rs#L191-L195

NyxCode:
So there's something left to fix (besides adding tests) after all.  
Anyway, @escritorio-gustavo, I pushed the changes here. Sorry for the one huge commit, though I added comments where I thought they might help. Let me know if anything is unclear.

gustavo-shigueo:
Don't worry about it! I've already left work for today, so I'll take a good look at it tomorrow

gustavo-shigueo:
> Not only are the import names broken

I remember implementing some kind of hack to try and deal with that in #235, I'll check what I was and make it less hacky

NyxCode:
> > Not only are the import names broken
> 
> I remember implementing some kind of hack to try and deal with that in #235, I'll check what I was and make it less hacky

I've seen that! Seemed like a reasonable solution, though we could also add a `const IDENT: &'static str` or `ident() -> String` to `TS` to return the name without generics.  
However, we're still using the `TypeId` for de-duping, and that changes if different generics are used.

NyxCode:
Ohh, I'm blind! I thought that was a `BtreeSet`, but we're collecting into a map, so we're just de-duping the names. Okay, easy fix then!

ts-rs/src/lib.rs
```diff
@@ -289,18 +288,25 @@ pub trait TS {
...
         panic!("{} cannot be declared", Self::name());
     }
 
+    /// Declaration of this type using the supplied generic arguments.
+    /// The resulting TypeScript definition will not be generic. For that, see `TS::decl()`.
+    /// If this type is not generic, then this function is equivalent to `TS::decl()`.
+    fn decl_concrete() -> String {
+        panic!("{} cannot be declared", Self::name());
+    }
+
```
NyxCode, line 306: Here's the user-visible effect of this change.
For
```rust
struct Generic<A, B> { a: A, b: B }
```,
`Generic::<i32, String>::decl_concrete()` returns `type Generic = { a: i32, b: String };`. 
`Generic::<i32, String>::decl()` swaps out the generic parameters `i32` and `String` out for on-the-fly generated dummy types `A` and `B`, resulting in `type Generic<A, B> = { a: A, b: B };`

File: macros/src/lib.rs [hunk 2]
```diff
@@ -131,10 +89,151 @@ impl DerivedTS {
...
+                        write!(f, "{:?}", self)
+                    }
+                }
+                impl TS for #generics {
+                    fn name() -> String { stringify!(#generics).to_owned() }
+                    fn transparent() -> bool { false }
+                }
+            )*
+        }
+    }
```
NyxCode, line 153: Here, these dummy types are generated, named the same as the generic parameters.  
I thought about only defining one dummy type, so e.g `#[docs(hidden)] pub struct Dummy<const NAME: &'static str>`.  
However, const-generics don't work with strings yet. Maybe something for the future to improve compile times.

macros/src/lib.rs
```diff
@@ -131,10 +89,151 @@ impl DerivedTS {
...
+            }
+            fn decl() -> String {
+                #generic_types
+                let inline = <#rust_ty<#(#generic_idents,)*> as ts_rs::TS>::inline();
+                let generics = #ts_generics;
+                format!("type {}{generics} = {inline};", #name)
+            }
+        }
+    }
 }
```
NyxCode, line 233: And finally, `decl` and `decl_concrete` are implemented here.  
`decl` first creates the dummy types, shadowing the names of the generics (`#generic_types`). Then, I just call `inline` on that type.

macros/src/types/enum.rs
```diff
@@ -124,9 +122,13 @@ fn format_variant(
                         (Some(_), Some(_)) => syn_err!("`type` is not compatible with `as`"),
                         (Some(type_override), None) => quote! { #type_override },
                         (None, Some(type_as)) => {
-                            format_type(&syn::parse_str::<Type>(&type_as)?, dependencies, generics)
+                            let ty = syn::parse_str::<Type>(&type_as)?;
+                            quote!(<#ty as ts_rs::TS>::name())
```
NyxCode, line 126: This is just one example of how every use of `generics::format_type` got replaced with `TS::name`.

escritorio-gustavo:
@NyxCode are we sure this fixes #214? While common traits such as `Debug` or `ToString` will work, I think user defined traits still fail right?

escritorio-gustavo:
This for example, fails to compile
```rs
trait TestingTrait {}

#[derive(TS)]
struct Testing<T: TestingTrait> {
    foo: T
}
```

escritorio-gustavo:
image

I tried adding `impl<T: TestingTrait> TestingTrait for Testing<T> {}` and the error persists

image

I suspect this is because the dummy type in `decl` doesn't implement `TestingTrait`, since there is no reasonable way to do that

NyxCode:
> @NyxCode are we sure this fixes #214? While common traits such as `Debug` or `ToString` will work, I think user defined traits still fail right?

You're right, this doesn't fix #214 in the general case. The test passes, since it didn't use any custom traits as trait bounds, so I enabled that.

NyxCode:
> I suspect this is because the dummy type in decl doesn't implement TestingTrait, since there is no reasonable way to do that
Right!
I think we have a way forward though: Once we get `&'static str` in const generics, we can replace the on-the-fly generated dummy types with just one dummy type `pub struct Dummy<const NAME: &'static str>`. Then, users could implement their traits for this struct.

escritorio-gustavo:
Great then in `decl` we'd have something like `Generic::<Dummy<"T">>::inline`

NyxCode:
Exactly! We could probably do something like that now, without const-generic `&'static str` by having `Dummy` by generic over some trait which provides the name instead. That'd give us #214, but won't improve compile times since we'd still need to generate types to put into that generic type.  
So idk, I feel like const-generic `&'static str` is something worth waiting for instead.

macros/src/lib.rs
```diff
@@ -131,10 +89,151 @@ impl DerivedTS {
...
+        }
+    }
+
+    fn generate_export_test(&self, rust_ty: &Ident, generics: &Generics) -> Option<TokenStream> {
+        let test_fn = format_ident!("export_bindings_{}", rust_ty.to_string().to_lowercase());
+        let generic_params = generics
+            .params
+            .iter()
+            .filter(|param| matches!(param, GenericParam::Type(_)))
+            .map(|_| quote! { () });
```
escritorio-gustavo, line 161: This here means we still using `Struct::<()>` for `TS::export` instead of the new dummy type, so
```rs
#[derive(TS)]
#[ts(export, export_to = "./tests-out/")]
struct Testing<T: ToString> {
    foo: T
}
```

Fails to compile when using `#[ts(export)]`

escritorio-gustavo:
> So idk, I feel like const-generic `&'static str` is something worth waiting for instead.

Cool! Do you have the link to the rust issue so we can track when this is added?

macros/src/lib.rs
NyxCode: Ah, shit! Yeah, you're right! That calls `Testing::<()>::decl()` and only then we swap it over to `Testing::<DUMMY>`.  
Ideally, we'd use the dummy types right away, but i'm not sure how to do this nicely.  
There are a couple alternatives to make this work, but I don't really like any of them
- Give users the ability to chose the generics
  - very unintuitive, since we swap them out anyway
- Generate the dummy types twice, once in `decl`, and onces in the export test
  - Even more code gen, not sure that's great for compile times
- Define one dummy type, and use that instead of `()` in the snippet you linked
  - Works (we swap it out anyway later), and doesn't generate more code than now, but it's a bit akward

macros/src/lib.rs
```diff
@@ -96,13 +96,8 @@ impl DerivedTS {
...
         let mut generics_ts_names = self
             .generics
-            .params
-            .iter()
-            .filter_map(|g| match g {
-                GenericParam::Lifetime(_) => None,
-                GenericParam::Type(ty) => Some(&ty.ident),
-                GenericParam::Const(_) => None,
-            })
+            .type_params()
```
NyxCode, line 99: Why didn't any1 tell me about this?? 😆 nice!

NyxCode:
> > So idk, I feel like const-generic `&'static str` is something worth waiting for instead.
> 
> Cool! Do you have the link to the rust issue so we can track when this is added?

https://github.com/rust-lang/rust/issues/95174
currently available in nightly using `#![feature(adt_const_params)]`

macros/src/lib.rs
escritorio-gustavo: > - Give users the ability to choose the generics
>    - very unintuitive, since we swap them out anyway

This is what I tried in #217, and it was pretty clunky

> - Generate the dummy types twice, once in decl, and onces in the export test
>    - Even more code gen, not sure that's great for compile times

Probably the easiest one to implement

> - Define one dummy type, and use that instead of () in the snippet you linked
>    - Works (we swap it out anyway later), and doesn't generate more code than now, but it's a bit akward

That might also be a pretty decent solution, and also would be the first step towards the single dummy type when `const T: &'static str` drops

macros/src/lib.rs
```diff
@@ -67,6 +67,10 @@ impl DerivedTS {
             #impl_start {
                 const EXPORT_TO: Option<&'static str> = Some(#export_to);
 
+                fn ident() -> String {
+                    stringify!(#rust_ty).to_owned()
+                }
+
```
NyxCode, line 73: That's not sufficient, since it doesn't account for `#[ts(rename)]`. The "real" TS ident is in `DerivedTS`, I think we should use that.

macros/src/lib.rs
```diff
@@ -151,7 +151,7 @@ impl DerivedTS {
         let test_fn = format_ident!("export_bindings_{}", rust_ty.to_string().to_lowercase());
         let generic_params = generics
             .type_params()
-            .map(|_| quote! { () });
+            .map(|_| quote! { ts_rs::Dummy });
```
NyxCode, line 155: Perfect!

ts-rs/src/lib.rs
```diff
@@ -727,3 +727,22 @@ impl_primitives! {
 }
 #[rustfmt::skip]
 pub(crate) use impl_primitives;
+
+
+#[doc(hidden)]
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, Ord, PartialOrd)]
+pub struct Dummy;
+
+#[doc(hidden)]
```
NyxCode, line 736: Does `#[doc(hidden)]` do anything on the `impl`s?

macros/src/lib.rs
NyxCode: Awesome, thx for implementing it!

ts-rs/src/lib.rs
escritorio-gustavo: I have no idea 😆. Just didn't wanna risk it

escritorio-gustavo: I thought it might because docs.rs shows what traits a type implements, but since the type itself has `#[doc(hidden)]` it may be redundant

NyxCode:
Hey @escritorio-gustavo !
I'm not sure! I think code-wise, it's fine with your changes (thanks!).  
Ideally, I'd like some more tests to make sure everything still kinda works, especially the dependencies. But maybe that's something we can do after merging this, but before a new release.

escritorio-gustavo:
> I'd like some more tests to make sure everything still kinda works, especially the dependencies. But maybe that's something we can do after merging this, but before a new release.

Agreed. I'm thinking we could have a separate PR that
- adds `TS_DIR=tests-out` to our test job
- adds `#[ts(export)]` to pretty much every type created in our tests
- runs `tsc --no-emit` on `tests-out`

escritorio-gustavo:
Hmm... I think I'll have to unlink #232. Although this PR removes the `unreachable!` problem when not using `inline`, it doen't fix the second issue reported there (which, admitedly, is a duplicate of #168)

escritorio-gustavo:
It seems like `inline` only breaks dependencies for `Result`, `Vec`, `HashMap` and other handwritten `impl`s where `transparent` returns true. We could make these just return `Self::name` or panic saying they can't be inlined

This idea seems ridiculous, but think about it:
`Vec<T>::name` returns `Array<T::name()>`
`Vec<T>::inline` returns `Array<T::inline()>`

But notice that `Vec<T>::inline` doesn't actually  inline `Vec`, instead it inlines `T`.
Why? Because there is no way to inline `Array`.

The problem is instead of stating this to be illegal (or just the same as calling `name`), we recursively
inline T

escritorio-gustavo:
The same is true for `Option` and `Result`. `name` returns them already inlined! With any other enum, inlining gives the type union representation, while these two do that for `name`. There is no inlining an enum's type union form!

In fact, if `Option` were exported, `name` would return `Option<T::name()>` and `inline` would return what `name` currently returns

escritorio-gustavo:
TL;DR

I think we should change the `impl`s in `ts-rs/lib.rs` because:
`Vec<T>::inline`, `HashMap<K, V>::inline`, `[T; N]::inline` can't exist, since there is no inlining for `Array<T>`,  `Record<K, V>` or `[T, T, ...]`, so `inline` should just defer to `name`
`Option<T>::name` and `Result<T, E>::name` either can't exist, since `Result` and `Option` aren't exported so `name` should defer to `inline`

escritorio-gustavo:
Alternatively, to keep current behavior, we have to recursively inline until we find a non transparent type

escritorio-gustavo:
@NyxCode I'll merge this PR and open a discussion for us to try and figure out how to best deal with all of these other issues

PR #221, opened 2024-02-01
[Proof of Concept] Typelist

NyxCode:

Issue #134, opened 2022-12-09
Use `{ [key: K]: V }` instead or `Record<K, V>` for Maps


ramiroaisen:
If we use 

```typescript
type T = Record<K, V>
```
and `V` is a type that circular reference type `T` we get an error in typescript.

This doesn't happen with 

```typescript
type T = { [key: K]: V }
``` 

If you are ok with this change I can make a PR, is a simple two liner change.

carlocorradini:
Off topic: How about enabling also [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)?

```ts
import type { B } from "./B";

export type A {
  map: Map<string, B>
}
```

escritorio-gustavo:
> and V is a type that circular reference type T we get an error in typescript.

Hey @ramiroaisen! Can you give an example of this for testing?

escritorio-gustavo:
> Off topic: How about enabling also [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)?

I don't think support for Map will be implemented, but you can use `#[type = "Map<K, V>"]`

escritorio-gustavo:
This issue seems to be stale, so I'll close it for now. If there is still interest in this feature, please open a new issue.

escritorio-gustavo:
Due to #276, this has become a relevant issue for new features, so I'm reopening the issue

PR #241, opened 2024-02-22
Bug: Type Parameters are not imported

NyxCode:


NyxCode:
While deriving `TS` for a struct, whenever we get to a field, we currently add that field to `Dependencies`.  
However, we additionally need to add all generic parameters of that field to `Dependencies` as well.

.github/workflows/test.yml
```diff
@@ -37,6 +37,24 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+      - name: example e2e test
+        working-directory: example
+        run: |
+          cargo t 
+          tsc bindings/* --noEmit
+      - name: workspace e2e with default export env
```
NyxCode, line 53: todo: missed that while copy-pasting

escritorio-gustavo:
>  I've implemented that fix here

Nice! I spent the last 30 minutes trying to get this to work before realizing there was the `TyoeList::push` method lol

```rs
fn generate_generics_fn(generics: &Generics) -> TokenStream {
    let generic_types = generics
        .type_params()
        .map(|ty| ty.ident.clone())
        .collect::<Vec<_>>();

    if generic_types.is_empty() {
        return quote!{
            fn generics() -> impl ts_rs::typelist::TypeList {}
        };
    }

    let generic_types = generic_types
        .iter()
        .fold(
            quote!(),
            |acc, cur| quote!{
                (std::marker::PhantomData<#cur>, #acc)
            }
        ); 
        
    quote!{
        fn generics() -> impl ts_rs::typelist::TypeList {
            #generic_types
        }
    }
}
```

NyxCode:
Alright, CI passes. (Meaning that our test coverage still isn't ideal ^^).

I haven't really touched the impls for `Result`, `Option`, etc., since they are, as you nicely outlined in the discussion, wonky.

escritorio-gustavo:
Great! At least our underlying problem with generics is fixed! Now we just have to decide on **how** to even handle the impls I mentioned in the discussion

escritorio-gustavo:
I think we have a new problem...

````rs
#[derive(TS)]
#[ts(export)]
struct Test {}

#[derive(TS)]
#[ts(export)]
enum MyOption<T> {
    Some(T),
    None
}

#[derive(TS)]
#[ts(export)]
struct Outer {
    a: MyOption<Test>
}
```

`Outer.ts` now generates correctly, but `MyOption.ts`
```ts
import type { Test } from "./Test";

export type MyOption<T> = { "Some": T } | "None";
```

It now imports `Test` for no reason

NyxCode:
> I think we have a new problem...
> 
> ```rust
> #[derive(TS)]
> #[ts(export)]
> struct Test {}
> 
> #[derive(TS)]
> #[ts(export)]
> enum MyOption<T> {
>     Some(T),
>     None
> }
> 
> #[derive(TS)]
> #[ts(export)]
> struct Outer {
>     a: MyOption<Test>
> }
> ```
> 
> `Outer.ts` now generates correctly, but `MyOption.ts`
> 
> ```ts
> // This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
> import type { Test } from "./Test";
> 
> export type MyOption<T> = { "Some": T } | "None";
> ```
> 
> It now imports `Test` for no reason

Ah :D
I think we had a similar issue before. Could you check if that issue is actually new?  
As far as I can tell, this is caused by both `Outer` and `MyOption` wanting to write `MyOption.ts`. Here, it seems like `Outer` won, and got to export `MyOption<Test>`, which depends on `Test`.

Gotta look into that a bit deeper. Definetely something we want to fix (I even enabled errors for unused imports in the CI), but at least it's not as bad as a missing import.

escritorio-gustavo:
It hink I found out why. The expansion of `Outer`'s `#[derive(TS)]` gives this:
```rs
    #[allow(clippy::unused_unit)]
    fn dependency_types() -> impl ts_rs::typelist::TypeList
    where
        Self: 'static,
    {
        {
            use ts_rs::typelist::TypeList;
            ().push::<MyOption<Test>>() // This should be MyOption<ts_rs::Dummy>
                .extend(<MyOption<Test> as ts_rs::TS>::generics())
        }
    }
```

I have copied the expansion and replaced the type manually, which fixed `MyOption.ts`

escritorio-gustavo:
> As far as I can tell, this is caused by both Outer and MyOption wanting to write MyOption.ts. Here, it seems like Outer won, and got to export MyOption<Test>, which depends on Test.

Yup, seems that way to me as well

> I think we had a similar issue before. Could you check if that issue is actually new?

Sure!

> It hink I found out why. The expansion of `Outer`'s `#[derive(TS)]` gives this:
> 
> ```rust
> #[allow(clippy::unused_unit)]
> fn dependency_types() -> impl ts_rs::typelist::TypeList
> where
>     Self: 'static,
> {
>     {
>         use ts_rs::typelist::TypeList;
>         ().push::<MyOption<Test>>() // This should be MyOption<ts_rs::Dummy>
>             .extend(<MyOption<Test> as ts_rs::TS>::generics())
>     }
> }
> ```
> 
> I have copied the expansion and replaced the type manually, which fixed `MyOption.ts`

If this really is the solution, I think type aliases are gonna be a problem again lol

NyxCode:
> If this really is the solution, I think type aliases are gonna be a problem again lol

True! I think we can do it a bit differently though:

Currently, we start at `Outer`, and recursively go through each dependency and export it. However, once we reach a type, `MyOption<Test>`, for example, we first generate the imports of `MyOption<Test>`, which includes `Test`.  
Only then do we call `MyOption<Test>::decl()`, which converts it into a `MyOption<SomeDummyType>`.  

If we had some way of first doing that conversion, and *then* doing the imports, we'd be fine

NyxCode:
One approach would be to make the conversion an operation independant of `TS::decl`. 

One way of doing this would be with an [associated type](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b77c7a39aae17ad10dd8a3a802c4f7a7). This is pretty ugly, especially since the dummy types would need to be in the public scope. With `const &'static str`, [that'd be much much nicer](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=64f949bd7b83131276af274fc0dee4fe)

Without `const &'static str`, we could [re-use the `TypeVisitor` trait](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ae5d278f6b55c32414b9d6060347ecef). Then, `decl()` would look like this:
```rs
fn decl() -> String {
    struct Visit(String);
    impl TypeVisitor for Visit {
        fn visit<T: TS>(&mut self) {
            self.0 = T::decl_concrete();
        }
    }
 
    let mut visit = Visit(String::new());
    Self::with_generic(&mut visit);
    visit.0
}
```

escritorio-gustavo:
> One approach would be to make the conversion an operation independant of `TS::decl`.
> 
> One way of doing this would be with an [associated type](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b77c7a39aae17ad10dd8a3a802c4f7a7). This is pretty ugly, especially since the dummy types would need to be in the public scope. With `const &'static str`, [that'd be much much nicer](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=64f949bd7b83131276af274fc0dee4fe)
> 
> Without `const &'static str`, we could [re-use the `TypeVisitor` trait](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ae5d278f6b55c32414b9d6060347ecef). Then, `decl()` would look like this:
> 
> ```rust
> fn decl() -> String {
>     struct Visit(String);
>     impl TypeVisitor for Visit {
>         fn visit<T: TS>(&mut self) {
>             self.0 = T::decl_concrete();
>         }
>     }
>  
>     let mut visit = Visit(String::new());
>     Self::with_generic(&mut visit);
>     visit.0
> }
> ```

Okay... I'm gonna be honest, I have no idea what any of that means lol.

> Alternatively, we could also do the import generation inside decl.

What would that look like?

escritorio-gustavo:
> One way of doing this would be with an [associated type](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b77c7a39aae17ad10dd8a3a802c4f7a7)

So if I got this right, we'd add `T::Generic` to `Dependencies` rather than just `T`?

escritorio-gustavo:
> I think we had a similar issue before. Could you check if that issue is actually new?

I couldn't find this error anywhere. The closest is #168, which brings extra imports to the type itself, but not its dependencies

NyxCode:
Just confirmed that the same thing happens on main, so it's not actually related to the changes in this PR

escritorio-gustavo:
Maybe we should rename `TS::generics` though... I managed to get myself confused between the associated type and the method with the same name lol

escritorio-gustavo:
I think this also needs to be done for `Range` and `RangeInclusive`?

NyxCode:
> @NyxCode this is my proposed fix for those types in the discussion. I have tested them against #168, even changing which types were `#[ts(inline)]`d and it seems to work as expected while keeping current behavior for `inline`

Awesome! I'll add a test for it!

NyxCode:
> I can also give the original idea (using the on-the-fly generated dummy types) a try to see how that'd look. Might be more complicated, but we'll see.

Played around with it a bit, but I'm running into `T: 'static` problems. So maybe that doesn't work after all.

NyxCode:
I renamed `TS::Generics` to `TS::WithoutGenerics`. Wasn't sure if `TS::WithoutGenerics` or `TS::WithDummyGenerics` was a better name, but I felt like `TS::WithoutGenerics` better conveyed the intention of the associated type.

escritorio-gustavo:
> Once we get `const S: &'static str`, `TS::WithoutGenerics` can actually be
> 
> ```rust
> struct Generic<A, B>(A, B);
> impl TS for Generic<A, B> {
>     type WithoutGenerics = Generic<Dummy<"A">, Dummy<"B">>;
>     // ...
> }
> ```
> 
> Then, we can re-write `TS::decl` to use that associated type, instead of generating new ones.

Hopefully this feature won't be in nightly forever like the `!` type lol. It would make these generics a lot simpler

macros/src/utils.rs
```diff
@@ -29,11 +29,16 @@ macro_rules! impl_parse {
             fn parse($input: syn::parse::ParseStream) -> syn::Result<Self> {
                 let mut $out = $i::default();
                 loop {
+                    let span = $input.span();
```
escritorio-gustavo, line 32: Getting the span before the key improves the location of the error:

## Assignment attributes
Before, points at the equals sign:
image

After, points at the attribute's name:
image

## Flag attributes:
Before, points at the token after the attribute (close bracket or comma):
image

After, points at the attribute's name:
image

macros/src/utils.rs
```diff
@@ -29,11 +29,16 @@ macro_rules! impl_parse {
...
                     let key: Ident = $input.call(syn::ext::IdentExt::parse_any)?;
                     match &*key.to_string() {
                         $($k => $e,)*
                         #[allow(unreachable_patterns)]
-                        _ => syn_err!($input.span(); "unexpected attribute")
+                        x => syn_err!(
+                            span;
+                            "Unknown attribute \"{x}\". Allowed attributes are: {}",
+                            [$(stringify!($k),)*].join(", ")
```
escritorio-gustavo, line 40: Tell the user what attributes are accepted, since LSP autocomplete can't do that

escritorio-gustavo:
Hey @NyxCode! Is anything else you think needs doing with this PR?

escritorio-gustavo:
Also one other thing I'd like to ask. Currently in this PR the only types where `TS::transparent()` returns `true` are `Range<I>` and `RangeInclusive<I>`. Do you think we should refactor these like what we did to `Result` and `Option`? This would allow us to remove `TS::transparent`, as it would always be `false`

macros/src/utils.rs
NyxCode: Love it, great work!

escritorio-gustavo:
> Hey @escritorio-gustavo! Hope you're doing well!

I'm doing great, thanks! Hope you are too

> Do you think it makes sense to include that in this PR?

I'm not really sure. One could argue it's an effect of the changes to `Result`, `Option`, `Vec` and `HashMap`. But there's also the problem of making this PR too big

> One thing I'd like to improve is the testing for dependencies. I think a while back you mentioned that we could, for example, just slap `#[ts(export)]` on every type in the test suite and run `tsc` over it

Yup! Some `#[ts(export)]` and maybe `#[ts(export_to = "...", rename = "...")]` to avoid naming collisions should allow `tsc` to check everything in the `tests` directory.

macros/src/utils.rs
escritorio-gustavo: Thanks! Btw, this is the first good look I've had at exactly how `impl_parse!` works, and I gotta say, I'm definetly using that for every proc_macro I make in the future lol. Awesome job man!

escritorio-gustavo:
> Yup! Some `#[ts(export)]` and maybe `#[ts(export_to = "...", rename = "...")]` to avoid naming collisions should allow `tsc` to check everything in the `tests` directory.

Minor update. I've already started doing this. I'm about halfway through. Should be able to commit the updated tests tomorrow

ts-rs/tests/raw_idents.rs
```diff
@@ -1,8 +1,10 @@
+#![allow(non_camel_case_types, dead_code)]
+
 use ts_rs::TS;
 
-#[allow(non_camel_case_types, dead_code)]
 #[derive(TS)]
-struct r#enum {
+#[ts(export, export_to = "tests-out/raw_idents/")]
+struct r#struct {
```
escritorio-gustavo, line 7: Had to rename `r#enum` to `r#struct` because `type enum = ...` is not valid TypeScript

ts-rs/tests/issue-70.rs
```diff
@@ -30,15 +32,25 @@ fn issue_70() {
     );
 }
 
+type GenericAlias<A = String, B = String> = HashMap<(A, String), Vec<(B, i32)>>;
```
escritorio-gustavo, line 35: This is gonna be a problem as `Record<K, V>` requires that `K extends string | number | symbol`

ts-rs/tests/issue-70.rs
```diff
@@ -30,15 +32,25 @@ fn issue_70() {
...
+
+#[derive(TS)]
+#[ts(export, export_to = "tests-out/issue_70/")]
+struct Container {
+    a: GenericAlias<Vec<i32>, Vec<String>>,
```
escritorio-gustavo, line 40: This is gonna be a problem as `Record<K, V>` requires that `K extends string | number | symbol`

ts-rs/tests/issue-70.rs
```diff
@@ -30,15 +32,25 @@ fn issue_70() {
...
+    b: GenericAlias,
+}
+
+#[derive(TS)]
+#[ts(export, export_to = "tests-out/issue_70/")]
+struct GenericContainer<A, B = i32> {
+    a: GenericAlias,
+    b: GenericAlias<A, B>,
+    c: GenericAlias<A, GenericAlias<A, B>>,
```
escritorio-gustavo, line 49: This is gonna be a problem as `Record<K, V>` requires that `K extends string | number | symbol`, for that to happen, our `A` generic would need to have that constraint (on top of changing the alias)

escritorio-gustavo (commented):
These issues may force us to implement #134

NyxCode:
> These issues may force us to implement https://github.com/Aleph-Alpha/ts-rs/issues/134

Hm. I think the idea here was to just come up with a very convoluted type to see if it generates correct TS.  
Neither `Record<K, V>` nor `{ [key: K]: V }` would support this (Afaik, they are identical). Same goes for `serde_json`, which just panics for a type like that.

So I'm not sure this is actually a bug. Maybe just changing the test to use a different type would be good enough.

escritorio-gustavo:
> Neither `Record<K, V>` nor `{ [key: K]: V }` would support this (Afaik, they are identical

You're right, the error just changes to be about index signatures being required to extend `string | number | symbol` or be a template literal type

escritorio-gustavo:
Okay, now that the updated tests are passing current CI, I found three tests that fail to generate correct imports:
- `generics.rs` -> types `Parent` and `GenericParent` missing an import to type `P0` (looks like some kind of depth limit to the imports)
- `lifetimes.rs` -> type `A` missing an import to type `B`
- `self_referential.rs` -> type `HasT` missing an import to type `T`

escritorio-gustavo:
> So I'm not sure this is actually a bug. Maybe just changing the test to use a different type would be good enough.

Agreed, however I don't think it's possible to fix `GenericContainer` unless we remove `HashMap` as we'd need `A extends string | number | symbol` to be able to use it in a `Record` as the key

escritorio-gustavo:
Okay, calling `std::any::type_name_of_val(&Parent::dependency_types)` gives the following TypeList:
```rs
(
      (
          (
              (
                  (
                      (
                          (),
                          (PhantomData::<T1<T0<u32>>>,)
                      ),
                      (
                          (),
                          (PhantomData::<T0<u32>>,)
                      )
                  ),
                  (PhantomData::<T1<P1<T0<P0<u32>>>>>,)
              ),
              (
                  (),
                  (PhantomData::<P1<T0<P0<u32>>>>,)
              )
          ),
          (PhantomData::<T1<P1<()>>>,)
      ),
      (
          (),
          (PhantomData::<P1<()>>,)
      )
  )
```

If I'm getting this right, the problem seems to be that `PhantomData::<P0<...>>` never shows up here. Is that right?

escritorio-gustavo:
The macro expansion's `dependency_types` returns the following:
```rs
().push::<T1<T0<u32>>>()
    .extend(<T1<T0<u32>> as ts_rs::TS>::generics())
    .push::<T1<P1<T0<P0<u32>>>>>()
    .extend(<T1<P1<T0<P0<u32>>>> as ts_rs::TS>::generics())
    .push::<T1<P1<()>>>()
    .extend(<T1<P1<()>> as ts_rs::TS>::generics())
```

NyxCode:
Very interesting!
I simplified it down to 
```rust
struct Parent {
    a: T1<T0<P0<u32>>>,
}
```

`Parent` depends on `[T1<...>] + T1<T0<P0<u32>>>::generic_types()`.
`T1<T0<P0<u32>>>::generic_types()` yields just `[T0]`, so `Parent::dependency_types()` is `[T1<...>, T0<...>]`.

Seems like I completely overlooked that generics are "transitive". As far as I can tell, `SomeType<SomeGeneric>::generic_types()` should not return `[SomeGeneric] + SomeGeneric::generic_types()`, but it only returns `[SomeGeneric]` currently. 

I'll try implementing that fix to see what happens if we do that 😆

escritorio-gustavo:
I added `.extend(#ident::generics())`, which seems to have worked
```rs
fn generate_generics_fn(&self) -> TokenStream {
    let generics = self
        .generics
        .type_params()
        .map(|TypeParam { ident, .. }| quote![.push::<#ident>().extend(<#ident as ts_rs::TS>::generics())]);
    quote! {
        #[allow(clippy::unused_unit)]
        fn generics() -> impl ts_rs::typelist::TypeList
        where
            Self: 'static,
        {
            use ts_rs::typelist::TypeList;
            ()#(#generics)*
        }
    }
}
```

NyxCode:
@escritorio-gustavo Ah, you were faster than me 😆! Nice work! Super cool that we caught this by running tsc over everything (though it's insanely slow on my machine, wtf)

escritorio-gustavo:
New CI finally passing 🎉

escritorio-gustavo:
I just realized, we will have to document all of this in CHANGELOG.md lol

.github/workflows/test.yml
```diff
@@ -60,11 +78,11 @@ jobs:
...
-          args: --all-features --manifest-path ts-rs/Cargo.toml
\ No newline at end of file
+      - name: No features
+        run: |
+          cargo test --no-default-features
+          tsc ts-rs/tests-out/**/*.ts --noEmit --noUnusedLocals
+      - name: All features
+        run: |
+          cargo test --all-features
+          tsc ts-rs/tests-out/**/*.ts --noEmit --noUnusedLocals
```
NyxCode, line 88: Turns out this doesn't work as-is. `/**/` Seems to only match one subdirectory, but it isn't recursive by default.  
I assume this is running on bash, where it can be enabled with `shopt -s globstar`. Commit is on the way.

escritorio-gustavo:
Okay, CI now fails, but it looks like it's just because TS hates self referential types

escritorio-gustavo:
Mine too!

escritorio-gustavo:
Also, notice that the push triggered test failed, but the PR triggered on succeeded

NyxCode:
Ah! Many types in `self_referential.rs` are renamed to `T`, so one would overwrite the other. In some cases one type might win over the other.

escritorio-gustavo:
Also, according to https://github.com/microsoft/TypeScript/issues/14174, it is supposed to work. Maybe the CI version of tsc is outdated?

NyxCode:
Now we actually see the real error since the files dont get overwritten! Seems like internally tagged enums don't like self-references.

escritorio-gustavo:
Looks like it doesn't like self references with type intesection

escritorio-gustavo:
> Not sure we can do anything about that.

Agreed. Self references are only allowed as property values

NyxCode:
Oh, look at that! Someone left a comment about this xD
> // NOTE: The generated type is actually not valid TS here, since the indirections rust enforces for recursive types
//       gets lost during the translation to TypeScript (e.g "Box<T>" => "T").

If that was me, then I dont remember ^^

I'll remove the `#[ts(export)]` then. I think it makes sense keeping the test itself though.

escritorio-gustavo:
This either never worked (as in the TS generated was always invalid), or it is a regression caused by enum flattening support

escritorio-gustavo:
> Oh, look at that! Someone left a comment about this xD
> 
> > // NOTE: The generated type is actually not valid TS here, since the indirections rust enforces for recursive types
> > //       gets lost during the translation to TypeScript (e.g "Box" => "T").

Well that literally explains a lot xD

> If that was me, then I dont remember ^^

Git blame says it was you 😆

NyxCode:
> This either never worked (as in the TS generated was always invalid), or it is a regression caused by enum flattening support

Okay, I did some thinking. This never worked in the first place, because it cant!

For one, the internally-tagged representation doesn't allow newtype/tuple variants. Serde just panics at runtime for them.  
That restriction makes sense, since having an internal tag requires the variant to be an object. 

"Internally tagged" means "put the tag inside the variant". But if the variant is not an object to put the tag in, this cant work.

Sensible types, like this one here work, since they don't use the type intersection:
```
#[derive(TS)]
#[ts(tag = "tag")]
enum X {
    A { inner: Box<Self> },
    B { inner: i32 }
}
```
```ts
export type X = { "tag": "A", inner: X, } | { "tag": "B", inner: number, };
```

escritorio-gustavo:
That makes a lot of sense! Especially given that one of the variants was trying to do
```ts
{ "tag": "E" } & Array<I>
```
which is just nonsense

escritorio-gustavo:
> For one, the internally-tagged representation doesn't allow newtype/tuple variants. Serde just panics at runtime for them.
That restriction makes sense, since having an internal tag requires the variant to be an object.

Added a note on the wiki about that

escritorio-gustavo:
I think it's because `impl_shadow` doesn't have `TS::generics`

NyxCode:
Yep! Seems like I just skipped that (since it didnt break any tests 😆)

escritorio-gustavo:
That's odd, in my computer TS complains that `CustomKey`(`null`) doesn't extend `string | number | symbol`, but CI still passes

escritorio-gustavo:
I'll be out for a about 2 hours now, so I'll be unable to respond. See you later

NyxCode:
> That's odd, in my computer TS complains that CustomKey(null) doesn't extend string | number | symbol, but CI still passes

Interesting! Seems like that's only caught with `--strict`, while `Record<{x: number}, Something>` also fails without `--strict`.  
I've enabled `--strict` for everything in CI, and fixed `CustomKey`. Everything works(TM) again!

---

I think it'd be alright if we merge this. It's getting annoying to scroll through this huge PR.  
Is there anything you think still needs to be done here?

Before the next release, I'd like to
- see if we can remove `TS::transparent()`
- Maybe make the recusive exporting of dependencies we added in #221 configurable?
  - While I can't think of a use-case where this would be undesirable, I think it's the change most likely to break existing setups.
    - Are there any other features you can think of we might want to make configurable for that reason? 

---

Thank you so much for your work, it has been invaluable! I'm super stoked to see where this library is going.  
Stuff just works(tm) now, especially inlining, other enum representations and generics. And now we've also increased the test suite by 160 `.ts` files. Just awesome stuff 😆

escritorio-gustavo:
> I think it'd be alright if we merge this. It's getting annoying to scroll through this huge PR.

Yup, this PR is way too big already

> Is there anything you think still needs to be done here?

While adding `#[ts(export)]` to everything, I noticed we don't have a `#[ts(...)]` equivalent to `#[serde(tag = "...")]` for structs, so I'd like to add that in a new PR.

> - see if we can remove TS::transparent()

Definetly something I'd like to see!

> - Maybe make the recusive exporting of dependencies we added in #221 configurable?

Cool! Would that be a feature flag or a new attribute?

>  - I think it's the change most likely to break existing setups.
>     - Are there any other features you can think of we might want to make configurable for that reason?

I don't see much of a problem, as long as the next release is `8.0.0` to indicate there are breaking changes.

---
> Thank you so much for your work, it has been invaluable! I'm super stoked to see where this library is going.

Thanks! It's awesome to be able to help out with a project that I use a lot!

PR #243, opened 2024-02-29
Remove `TS::trasparent`

escritorio-gustavo:
This PR is meant to remove the `TS::transparent` method, as it was rendered useless by #241, always return `false` with the exception of `Range` and `RangeInclusive`

### DIFF
--- a/ts-rs/src/lib.rs
+++ b/ts-rs/src/lib.rs
@@ -502,6 +502,7 @@ macro_rules! impl_shadow {
     (as $s:ty: $($impl:tt)*) => {
         $($impl)* {
             type WithoutGenerics = <$s as TS>::WithoutGenerics;
+            fn ident() -> String { <$s>::ident() }
             fn name() -> String { <$s>::name() }
             fn inline() -> String { <$s>::inline() }
             fn inline_flattened() -> String { <$s>::inline_flattened() }
--- a/ts-rs/src/lib.rs
+++ b/ts-rs/src/lib.rs
@@ -367,12 +367,7 @@ pub trait TS {
         struct Visit<'a>(&'a mut Vec<Dependency>);
         impl<'a> TypeVisitor for Visit<'a> {
             fn visit<T: TS + 'static + ?Sized>(&mut self) {
-                if T::transparent() {
-                    // the dependency `T` is "transparent", meaning that our original type depends
-                    // on the dependencies of `T` as well.
-                    T::dependency_types().for_each(self);
-                } else if let Some(dep) = Dependency::from_ty::<T>() {
-                    // the dependency `T` is not transparent, so we just add it to the output
+                if let Some(dep) = Dependency::from_ty::<T>() {
                     self.0.push(dep);
                 }
             }
@@ -381,10 +376,6 @@ pub trait TS {
 
         deps
     }
-
-    /// `true` if this is a transparent type, e.g tuples or a list.
-    /// This is used for resolving imports when using the `export!` macro.
-    fn transparent() -> bool;
 
     /// Manually export this type to a file.
     /// The output file can be specified by annotating the type with `#[ts(export_to = ".."]`.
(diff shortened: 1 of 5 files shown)


ts-rs/src/lib.rs
```diff
@@ -705,11 +676,15 @@ impl<I: TS> TS for Range<I> {
...
-    fn transparent() -> bool {
-        true
+    fn generics() -> impl TypeList
+    where
+        Self: 'static,
+    {
+        use std::marker::PhantomData;
+        ((PhantomData::<I>,), I::generics())
     }
 }
```
NyxCode, line 689: ~~So, this is equivalent to just returning `I::generics()`.  
I'm not sure that's correct, even though the tests pass. 
Shouldn't this return `().push::<I>().extend(I::generics())`? Right now, `I` itself would be missing from the TypeList, no?~~

Ahhh, I got confused by you constructing the TypeList manually here!  
Is there a reason you're not doing `().push::<I>().extend(I::generics())` or, even better, `I::generics().push::<I>`, instead?

ts-rs/src/lib.rs
```diff
@@ -367,12 +367,7 @@ pub trait TS {
...
             fn visit<T: TS + 'static + ?Sized>(&mut self) {
-                if T::transparent() {
-                    // the dependency `T` is "transparent", meaning that our original type depends
-                    // on the dependencies of `T` as well.
-                    T::dependency_types().for_each(self);
-                } else if let Some(dep) = Dependency::from_ty::<T>() {
-                    // the dependency `T` is not transparent, so we just add it to the output
+                if let Some(dep) = Dependency::from_ty::<T>() {
                     self.0.push(dep);
                 }
```
NyxCode, line 372: I love that this just works. Awesome!

NyxCode (commented):
Wow, this was much much more straight-forward than I thought. Awesome!

ts-rs/src/lib.rs
escritorio-gustavo: Not really, I just kinda got confused calling methods on `()` so I did it this way

ts-rs/src/lib.rs
escritorio-gustavo: Yeah, since `T::transparent` is always `false` we just get rid of that check

ts-rs/src/lib.rs
escritorio-gustavo: > I just kinda got confused calling methods on `()` so I did it this way

That tripped me up quite a bit in the macros, because the repetition syntax uses parentheses plus the unit type plus the calling of methods

NyxCode: Hm, yeah, I understand.  
I do think it'd be a good idea to keep the implementation of `TypeList` private, and use the methods on it. That gives us the option to change TypeList is implemented however we like in the future.

In this particular case, `I::generics().push::<I>()` would probably be the nicest.  
To make the code more readable (and keep the implementation detail that the empty TypeList is `()` hidden), we could add a `const EMPTY_TYPELIST: () = ()` or `pub fn new() -> impl TypeList {}` to `typelist.rs`

For the macros, I try to use `quote!{}` or `quote![]` to disambiguate them, but yeah, there's still the `#()*` syntax.

In the end, this is not a big deal, if you've got strong opinions here. Personally, I prefer using the methods on the type.

Edit: I see that you already changed it, thanks! 😆 

Being able to call methods on `TypeList` was the motivation to make `dependency_types` a function in the first place. It could have been an associated type (e.g `type Dependencies = ((A), B), C::Dependencies)`), but I liked the idea of being able to pretend that `TypeList` is just a `Vec`.

NyxCode (approved):

ts-rs/src/lib.rs
escritorio-gustavo: All of that makes a lot of sense! I'll open another PR to apply this change to the impls I changed in #241, all of which manually build a TypeList

PR #244, opened 2024-02-29
Parse `#[ts(tag = "...")]` on structs

escritorio-gustavo:
This PR adds a `#[ts(...)]` equivalent to `#[serde(tag = "...")]` on structs


NyxCode (approved):
Ohh, right, we've already got `tag` on `StructAttr` since we use that for enums under the hood. Very elegant, love it!

escritorio-gustavo:
> Ohh, right, we've already got tag on StructAttr since we use that for enums under the hood

I don't think that's why we already had `tag` on `StructAttr`, since `EnumAttr` has its own `tag` to handle that. Instead we did already support the serde feature of adding a tag to a struct, but it required `serde-compat` and `#[serde(...)]`

PR #246, opened 2024-02-29
No longer build TypeLists manually

escritorio-gustavo:
As discussed in #243, typelists should not be manually built


NyxCode (approved):
Awesome, thanks!

PR #247, opened 2024-02-29
bug: `#[ts(export_to = "../path")]` can cause `diff_paths` to fail

escritorio-gustavo:
This PR reproduces the potential for paths containing `..` to break `diff_paths`

### DIFF
--- a/ts-rs/src/export.rs
+++ b/ts-rs/src/export.rs
@@ -6,6 +6,7 @@ use std::{
     sync::{Mutex, OnceLock},
 };
 
+use path_clean::PathClean;
 use thiserror::Error;
 use ExportError::*;
 
@@ -221,6 +222,14 @@ fn import_path(from: &Path, import: &Path) -> String {
     }
 }
 
+fn to_absolute_path(path: &Path) -> PathBuf {
+    let Ok(current_path) = std::env::current_dir() else {
+        return path.to_owned()
+    };
+
+    current_path.join(path).clean()
+}
+
 // Construct a relative path from a provided base directory path to the provided path.
 //
 // Copyright 2012-2015 The Rust Project Developers.
@@ -238,12 +247,12 @@ where
     P: AsRef<Path>,
     B: AsRef<Path>,
 {
-    let path = path.as_ref();
-    let base = base.as_ref();
+    let path = to_absolute_path(path.as_ref());
+    let base = to_absolute_path(base.as_ref());
 
     if path.is_absolute() != base.is_absolute() {
         if path.is_absolute() {
-            Some(PathBuf::from(path))
+            Some(path)
         } else {
             None
         }
--- a/ts-rs/src/export.rs
+++ b/ts-rs/src/export.rs
@@ -6,7 +6,6 @@ use std::{
     sync::{Mutex, OnceLock},
 };
 
-use path_clean::PathClean;
 use thiserror::Error;
 use ExportError::*;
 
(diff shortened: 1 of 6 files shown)


ts-rs/src/export.rs
```diff
@@ -221,6 +222,14 @@ fn import_path(from: &Path, import: &Path) -> String {
     }
 }
 
+fn to_absolute_path(path: &Path) -> PathBuf {
+    let Ok(current_path) = std::env::current_dir() else {
+        return path.to_owned()
+    };
+
+    current_path.join(path).clean()
```
NyxCode, line 230: I think we should just copy [these 16](https://docs.rs/path-clean/latest/src/path_clean/lib.rs.html#62-78) lines into here.

ts-rs/src/export.rs
```diff
@@ -221,6 +222,14 @@ fn import_path(from: &Path, import: &Path) -> String {
     }
 }
 
+fn to_absolute_path(path: &Path) -> PathBuf {
+    let Ok(current_path) = std::env::current_dir() else {
+        return path.to_owned()
+    };
```
NyxCode, line 228: I think we can safely unwrap this.

ts-rs/src/export.rs
```diff
@@ -238,12 +247,12 @@ where
...
+    let path = to_absolute_path(path.as_ref());
+    let base = to_absolute_path(base.as_ref());
 
     if path.is_absolute() != base.is_absolute() {
         if path.is_absolute() {
-            Some(PathBuf::from(path))
+            Some(path)
         } else {
             None
         }
```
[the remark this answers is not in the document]
escritorio-gustavo: That's true, but if `std::env::current_dir` errors, the path is returned as is

ts-rs/src/export.rs
escritorio-gustavo: I see... Nevermind my previous comment then xD

escritorio-gustavo: I would still like to change diff_paths to return a `Result` and use `?` instead of `unwrap`ping though

ts-rs/src/export.rs
escritorio-gustavo: Fair enough, that avoids a dependency

escritorio-gustavo: The `diff_paths` function has a comment about licensing. Should I copy it over to this as well, `path-clean` uses the same license described there (APACHE or MIT)

escritorio-gustavo: Alright, I did end up adding the licensing info. But I'm also starting to think the `export.rs` file is doing a lot more than it should. Do you mind if I break some of it into other modules?

NyxCode:
I think we can dumb it down a bit, [something like this](https://gist.github.com/NyxCode/bc0d2ac1b0af06a98e4795017926937b).  
That inlines `clean_path` and uses the fact that we know that the input to `clean_path` starts at the root to simplify it.

escritorio-gustavo:
Oh, I have kinda been doing the opposite lol

I separated absolute and clean into their own modules and crates. I haven't pushed this commit yet, so I can totally just reverse it and replace with the gist you sent

escritorio-gustavo:
Another thing I did is change `current_dir` to `CARGO_MANIFEST_DIR` as you suggested. They are always the same as far as ts-rs is concerned and `CARGO_MANIFEST_DIR` is already used elsewhere, so it would be inconsistent to use `current_dir`.

I also used the `std::env!` macro instead of `std::env::var().unwrap()`

escritorio-gustavo:
Hey, sorry I vanished, had to do some stuff at work, so it may take a few hours before I'm able to make more changes.

I do still have the commit I mentioned, but as I said it goes pretty much in the opposite direction you asked for. Do you want me to push it anyway so you can have a look?

escritorio-gustavo:
> Do you want me to push it anyway so you can have a look?

Hey @NyxCode, I did push it as a PR into this branch #248. I'm gonna have to leave again for a few hours. If you don't like it I will close #248 and apply the gist instead.

escritorio-gustavo:
Hey @NyxCode, check out this version, it does inline and simplify the cleaning of the path like in the gist you submitted. It also renames the function to `absolute` and moves it into a `mod` called `path`, so if `std::path::absolute` ever comes to stable Rust, we can just add `std::` to these calls

NyxCode:
Hey! Sorry for not responding sooner, still busy atm. Will probably stay that way until march 28th.

This PR looks great! Doesn't overcomplicate anything while keeping `export.rs` tidy.  
Only thing that caught my eye is that the test should probably have an `assert!(Path::new("..").is_file())` to make sure everything is actually in the right place.  
Besides that, this looks nice, thanks! Feel free to merge it.

NyxCode (approved):

escritorio-gustavo:
> Hey! Sorry for not responding sooner, still busy atm. Will probably stay that way until march 28th.

No problem, take your time!

> Only thing that caught my eye is that the test should probably have an `assert!(Path::new("..").is_file())` to make sure everything is actually in the right place.

I'm a bit confused, which test should have that assert? Also, doesn't `Path::new("..").is_file()` evaluate to `false`?

NyxCode:
Could have worded that better, sry about that 😆   
From my side, we can merge this. Nice work.

PR #248, opened 2024-03-01
Path bug proposed fix

escritorio-gustavo:


escritorio-gustavo:
@NyxCode, It seems like there is a difference between `CARGO_MANIFEST_DIR` and `current_dir`

`CARGO_MANIFEST_DIR` causes `example` to export into `ts-rs/bindings`

escritorio-gustavo:
After messing about with this a bit more, I figured out how to make `#[ts(export_to = "...")]` relative to `TS_RS_EXPORT_DIR` when they are both being used. Also managed to get rid of the `__private` module and `TS::get_export_to` function, going back to our ol'reliable `TS::EXPORT_TO`

PR #250, opened 2024-03-06
Make `#[ts(export_to = "...")]` relative to `TS_RS_EXPORT_DIR`

escritorio-gustavo:
As discussed in #245, when using both `TS_RS_EXPORT_DIR` and `#[ts(export_to = "...")]`, the `export_to` path should be relative to `TS_RS_EXPORT_DIR` instead of the project's `Cargo.toml`

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -34,43 +34,11 @@ impl DerivedTS {
             .export
             .then(|| self.generate_export_test(&rust_ty, &generics));
 
-        let export_dir = std::option_env!("TS_RS_EXPORT_DIR");
         let export_to = {
             let path = match self.export_to.as_deref() {
-                Some(dirname) if dirname.ends_with('/') => {
-                    export_dir
-                        .map_or_else(
-                            || format!("{}{}.ts", dirname, self.ts_name),
-                            |ts_rs_dir| format!(
-                                "{}/{}{}.ts",
-                                ts_rs_dir.trim_end_matches('/'),
-                                dirname,
-                                self.ts_name,
-                            )
-                        )
-                },
-                Some(filename) => {
-                    export_dir
-                        .map_or_else(
-                            || filename.to_owned(),
-                            |ts_rs_dir| format!(
-                                "{}/{}",
-                                ts_rs_dir.trim_end_matches('/'),
-                                filename,
-                            ),
-                        )
-                },
-                None => {
-                    export_dir
-                        .map_or_else(
-                            || format!("bindings/{}.ts", self.ts_name),
-                            |ts_rs_dir| format!(
-                                "{}/{}.ts",
-                                ts_rs_dir.trim_end_matches('/'),
-                                self.ts_name,
-                            ),
-                        )
-                },
+                Some(dirname) if dirname.ends_with('/') => format!("{}{}.ts", dirname, self.ts_name),
+                Some(filename) => filename.to_owned(),
+                None => format!("bindings/{}.ts", self.ts_name),
             };
 
             quote! {
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -30,25 +30,53 @@ struct DerivedTS {
 
 impl DerivedTS {
     fn into_impl(mut self, rust_ty: Ident, generics: Generics) -> TokenStream {
-        let mut get_export_to = quote! {};
-        let export_to = match &self.export_to {
-            Some(dirname) if dirname.ends_with('/') => {
-                format!("{}{}.ts", dirname, self.ts_name)
-            }
-            Some(filename) => filename.clone(),
-            None => {
-                get_export_to = quote! {
-                    fn get_export_to() -> Option<String> {
-                        ts_rs::__private::get_export_to_path::<Self>()
-                    }
-                };
-                format!("bindings/{}.ts", self.ts_name)
-            }
-        };
-
         let export = self
             .export
             .then(|| self.generate_export_test(&rust_ty, &generics));
+
+        let export_dir = std::env::var("TS_RS_EXPORT_DIR").ok();
+        let export_to = {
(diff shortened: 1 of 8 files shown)


escritorio-gustavo:
I pushed the wrong branch... sorry

escritorio-gustavo:
This is so annoying, docs.rs passes on my computer, but not on ci

escritorio-gustavo:
It appears that for some reason the path on ci contains one extra `ts-rs`

NyxCode:
I don't quite understand what's going on, but it seems like `TS_RS_EXPORT_DIR` is not respected.  

in `export_type`, we call `output_path`, but that doen't take `TS_RS_EXPORT_DIr` into account at all.

NyxCode:
Running `TS_RS_EXPORT_DIR=./custom cargo test -- export_bindings` still puts everything in `tests-out`

macros/src/lib.rs
```diff
@@ -30,26 +30,54 @@ struct DerivedTS {
...
-                };
-                format!("bindings/{}.ts", self.ts_name)
-            }
-        };
-
         let export = self
             .export
             .then(|| self.generate_export_test(&rust_ty, &generics));
 
+        let export_dir = std::env::var("TS_RS_EXPORT_DIR").ok();
```
NyxCode, line 37: Ahh, I think this here might be the issue!  
This is evaluated when the macro executes, during compilation.  

When later running this executable, the `TS_RS_EXPORT_DIR` is hard-coded in there.  
That might also happen when just doing `cargo test`, since rust might not re-compile.

escritorio-gustavo: I think you're right, I just commited a version using `std::option_env!` instead and that seems to have fixed it

escritorio-gustavo: Still, it might be better to move reading the env variable to `export.rs` instead, because I think the way I just did it forces recompilation, which kinda sucks

NyxCode: Agreed, both with the observation as well as the conclusion.

NyxCode: The last change did cause a recompilation, but recompilation of `ts-rs-macros`.  
The 3 possible behaviours we could get are
- Re-compile ts-rs-macros 
  Use `env!` in `ts-rs-macros`
- Re-compile the consumer crate
  Load the environment variable in `export.rs`, using `env!()`
- Don't recompile anything
  Load the environment variable in `export.rs`, using `env::var()`

escritorio-gustavo:
I finished moving `std::env::var` outside of the macro. The only downside to this is that `T::EXPORT_TO` doesn't actually correspond to the exported path anymore, due to the lack of `TS_RS_EXPORT_DIR`, but this should reduce the amount of compilation going on in the user's machine

escritorio-gustavo:
I could add a method that returns the proper path to remedy this

macros/src/lib.rs
escritorio-gustavo: Got it fixed! Sorry about the billion failed CI notifications 😅

escritorio-gustavo: > * Don't recompile anything
>       Load the environment variable in `export.rs`, using `env::var()`

Ended up on this one, my cpu is already thanking me 😆

NyxCode:
LGTM! I'd be fine if you want to pull the "read env var or ".", get EXPORT_TO or T::ident, concat them" logic into a separate function to re-use that in `Dependency::from_ty` and `generate_imports`.  

Making it a provided method on `TS` might be a bit akward, since it should under no circumstances be overwritten, so maybe something to put in `export/path.rs`?

Doesn't have to be though if you dont think it makes sense. I'm just happy that we got rid of that weird `TS::get_export_to`, that was always confusing to me.

ts-rs/src/export.rs
```diff
@@ -181,9 +166,15 @@ fn generate_decl<T: TS + ?Sized>(out: &mut String) {
...
+        .ok()
+        .as_deref()
+        .map(Path::new)
+        .unwrap_or_else(|| Path::new("."))
+        .to_owned();
     let export_to =
-        T::get_export_to().ok_or(ExportError::CannotBeExported(std::any::type_name::<T>()))?;
-    let path = Path::new(&export_to);
+        T::EXPORT_TO.ok_or(ExportError::CannotBeExported(std::any::type_name::<T>()))?;
+    let path = base.join(export_to);
```
NyxCode, line 177: Could we re-use `output_path` here? Might be the case that we have to get rid of the `path::absolute` in `output_path` for that.

escritorio-gustavo:
The reason I was thinking about adding it to `TS` is because of the tests that read the exported file, all of them have a bunch of weird stuff to handle `TS_RS_EXPORT_DIR`

ts-rs/src/lib.rs
```diff
@@ -428,7 +421,17 @@ impl Dependency {
...
+            .ok()
+            .as_deref()
+            .map(Path::new)
+            .unwrap_or_else(|| Path::new("."))
+            .to_owned();
+
+        let exported_to = base
+            .join(T::EXPORT_TO.map(ToOwned::to_owned)?)
+            .to_string_lossy()
+            .to_string();
```
NyxCode, line 434: Same thing here, might be nice to re-use `output_path` here as well

NyxCode:
Yeah, that's a good point. If you think it makes sense, go for it!

ts-rs/src/export.rs
escritorio-gustavo: I don't even think we need to remove `path::absolute`, this path will be fed into `diff_paths`, which uses absolute paths as well

NyxCode: you're right, perfect!

escritorio-gustavo:
I agree that the function that returns the path should never be overwritten. Maybe a `pub fn` would be better than a new `TS` method then?

PR #252, opened 2024-03-06
Update & format documentation

NyxCode:


NyxCode:
Arghh. Docs look much nicer on using `cargo docs`, but now the README looks awefull.

escritorio-gustavo:
We have some methods that are provided but their default implementation just panics. Should we remove the default impl to show that these methods are required in case someone decides to manually implement `TS`?

NyxCode:
Thanks for cleaning everything up! <3

> Should we remove the default impl to show that these methods are required in case someone decides to manually implement TS?

Sure, I like that idea! Can we already merge this, since you opened a separate PR for that?

NyxCode:
Hm, actually, I think I'll add a few docs to `typelist`, that's entirely undocumented right now, and it is part of the public interface.

escritorio-gustavo:
> Can we already merge this, since you opened a separate PR for that?

Sure! It's all good from my side

PR #253, opened 2024-03-07
Ignore ts-rs tests by default

### LINKED ISSUE #62: Avoiding the testcase workaround
Hello,

as you're aware, hiding the code generation in testcases is a bit of a hack and probably annoying for anyone actually using their tests for other matters. Like testing.

If you haven't seen it, https://github.com/Wulf/tsync is taking a different approach: Instead of generating a meaningful `#[derive(TS)]` at build time, it will parse your source code independently of the compiler. But both tsync and ts-rs use the `syn` crate, and will parse a `syn::Item::Struct`/`::Enum` into type information.

Their approach has a few advantages:
  * doesn't mess with test cases
  * no build overhead to generate meaningful `#[derive(..)]` (though that is probably minor)
  * it's easier to regenerate .d.ts files on demand (just run tsync instead of all your tests)

But let's not hide the disadvantages:
  * they currently require global installation of a `tsync` binary. This could be avoided by letting users add a one-liner `[[bin]]` or `[[example]]` to their project and calling that. (an example has the advantage that you can put its deps into dev-dependencies)

I was wondering if it'd be useful to better abstract the type parsing away from the type consumption. We create a function taking a `syn::Item` and returning a `struct DerivedTS` (or something more generic than it).

Using that function, we could implement both the actual `#[derive(TS)]` and testcase generation, but we could also implement an external parser similar to tsync for projects where that's a better fit.

Do you think that'd be useful? Am I overlooking something important? Is this something I should hack on the next slow weekend, or over the holidays?

escritorio-gustavo:
Currently ts-rs exports its TypeScript bindings in every test run. This seems to really upset some people, and with good reasons:
- One cannot run their own tests without triggering ts-rs
- One cannot trigger ts-rs without running their own tests
- Both situations cause slowdowns due to running more tests than expected

So this PR proposes a way to change that by adding `#[ignore]` to all ts-rs tests. This enables users to:
- Run only their own tests with `cargo test`
- Run only ts-rs tests with `cargo test export_bindings_ -- --ignored`
- Run everything with `cargo test -- --ignored` (though this does have the downside of running the user's own ignored tests)

Fixes #62


escritorio-gustavo:
Obsolete as of #304

PR #254, opened 2024-03-07
Convert inline, inline_flattened, decl and decl_concrete to 

escritorio-gustavo:


NyxCode (approved):
Looks good, feel free to merge!

PR #256, opened 2024-03-07
Alternative handling of output paths

NyxCode:
Hey @escritorio-gustavo!

This is an experiment to see if we can further simplify and streamline the handling of export paths.  
Let's start with a practical example:

# Example
## Before
In the workspace E2E test, there's the `parent` crate, which depends on `crate1`.  
The struct `crate1::Crate1` has `#[ts(export_to = "crate1/")]`, and 
the struct `parent::Parent` has `#[ts(export)]`.

When running `cargo test`, only `parent::Parent` exports, but it takes its dependency `crate1::Crate1` with it.
That results in this directory structure:
```
parent/
    bindings/
        Parent.ts
    crate1/
        Crate1.ts
```

## After
With this change, the resulting directory structure looks like this:
```
parent/
    bindings/
        Parent.ts
        crate1/
            Crate1.ts
```

# Explanation
## Before
For a type, the output path is   
`./{TS_RS_EXPORT_DIR}/bindings/{name}.ts` if `#[ts(export_to = "...")]` is **not** specified and
`./{TS_RS_EXPORT_DIR}/{name}.ts` if `#[ts(export_to = "...")]` is specified.

This is because `TS::EXPORT_TO` defaults to `bindings/{name.ts}`, but changing it using `#[ts(export_to = "...")]` gets rid of the `bindings/` prefix.  
The environment variable `TS_RS_EXPORT_DIR` defaults to `.`.

## After
With this change, it's the environment variable `TS_RS_EXPORT_DIR` which defaults to `./bindings`.

Therefore, with this change, the output path is always  `./{TS_RS_EXPORT_DIR}/{name}.ts`.  
We take the `TS_RS_EXPORT_DIR` (or `./bindings`), append the name to it (or the `#[ts(export_to = "...")]` override), and we're done.

---

I feel like this behaviour better matches what users would expect. They specify where there files should go relative to some base directory, `TS_RS_EXPORT_DIR`.  
What do you think?


macros/src/lib.rs
```diff
@@ -34,17 +34,22 @@ impl DerivedTS {
...
             };
 
             quote! {
-                const EXPORT_TO: Option<&'static str> = Some(#path);
+                fn output_path() -> Option<std::path::PathBuf> {
+                    let path = std::env::var("TS_RS_EXPORT_DIR");
+                    let path = path.as_deref().unwrap_or("./bindings");
+
+                    Some(std::path::Path::new(path).join(#path))
+                }
```
NyxCode, line 52: The implementation of that is super simple - Take `TS_RS_EXPORT_DIR` (or `./bindings`), and append the name (or `#[ts(export_to = "...")]` override to it.

NyxCode:
In our example, `cargo test` does the same as before:  
```
example/
  bindings/
    ...
```
With `TS_RS_EXPORT_DIR=./typescript`, however, you'd get this:
```
example/
  typescript/
    bindings/
      ...
```

With this change, you'd get
```
example/
  typescript/
    ...
```

example/src/lib.rs
```diff
@@ -9,7 +9,7 @@ use uuid::Uuid;
 
 #[derive(Serialize, TS)]
 #[ts(rename_all = "lowercase")]
-#[ts(export, export_to = "bindings/UserRole.ts")]
+#[ts(export, export_to = "UserRole.ts")]
 enum Role {
     User,
     #[ts(rename = "administrator")]
```
NyxCode, line 15: Before, renaming the file from `Role.ts` to `UserRole.ts` required adding the `bindings/` prefix back again.  
With this change, this is no longer necessary.

escritorio-gustavo:
> For a type, the output path is
> `./{TS_RS_EXPORT_DIR}/bindings/{name}.ts` if `#[ts(export_to = "...")]` is not specified

Oof, this `bindings` is not supposed to be there, this is a major oversight from my part when doing #250, caused by the fact that all of our tests use `#[ts(export_to = "...")]`

escritorio-gustavo:
> ## After
> 
> With this change, it's the environment variable `TS_RS_EXPORT_DIR` which defaults to `./bindings`.
> 
> Therefore, with this change, the output path is always `./{TS_RS_EXPORT_DIR}/{name}.ts`. We take the `TS_RS_EXPORT_DIR` (or `./bindings`), append the name to it (or the `#[ts(export_to = "...")]` override), and we're done.

Very nice way of dealing with it!

escritorio-gustavo (approved):

NyxCode:
Alright, then I'll go ahead and merge this!  
It *is* a breaking change, but it's probably a good time to do it.  
All the tests, for example, have `#[ts(export_to = "tests_out/)]`, so they end up in `ts-rs/bindings/test-out`.

escritorio-gustavo:
> All the tests, for example, have `#[ts(export_to = "tests_out/")]`

We could make a separate PR to remove `tests-out` from the path to reduce nesting

PR #258, opened 2024-03-08
Add dependency caching to GH Actions for faster CI

escritorio-gustavo:
Our CI takes quite a while to run, mostly because of the Rust compiler being kinda slow. This PR attempts to remedy that by caching its dependencies. It also runs twice on every commit when we do PR's from within the repo, this PR also fixes that.

I have no idea if it will actually work, but should be worth the shot

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,15 +10,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
-      - uses: actions/cache@v3
-        with:
-          path: |
-            ~/.cargo/bin/
-            ~/.cargo/registry/index/
-            ~/.cargo/registry/cache/
-            ~/.cargo/git/db/
-            e2e/dependencies/consumer/target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('e2e/dependencies/consumer/Cargo.lock') }}
+      - uses: Swatinem/rust-cache@v2
       - name: dependencies e2e test
         working-directory: e2e/dependencies/consumer
         run: |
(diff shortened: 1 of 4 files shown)


escritorio-gustavo:
Okay, first CI run finished in 2m 30s.

We have a few dead code warnings in our tests, I'll add `#![allow(dead_code)]` to remove those and trigger CI so we can see if there is a difference

escritorio-gustavo:
2m 22s... that's disappointing

escritorio-gustavo:
I think it's not caching because it has to compile both `--all-features` and `--no-default-features`. I'll try splitting them up to see if helps, if it doesn't, I'll just close this PR

escritorio-gustavo:
Oops, running these tests in parallel breaks them due to everyone messing with the same files

escritorio-gustavo:
> Instead of running serially, stop CI from running twice

Insetead of this, I stopped CI from running at all smh

escritorio-gustavo:
This is not gonna work

escritorio-gustavo:
Hold on, it seems I still have a syntax error

escritorio-gustavo:
Well, I finally got CI to run again

escritorio-gustavo:
It is about 30-40 seconds faster, but recompilation is still happenning, I'll try changing `CARGO_HOME`

escritorio-gustavo:
Finally! Got the time down to 56s

escritorio-gustavo:
> Finally! Got the time down to 1m06s

The only reason for it to still be this slow is that the `--all-features` test still takes 40s+ to compile for some reason... still, better than the previous 2m+

NyxCode (approved):
> It also runs twice on every commit when we do PR's from within the repo, this PR also fixes that.

🥳

Great change, super happy to see this!  
To keep main clean, it'd be nice if you could squash the commits.

PR #262, opened 2024-03-10
Make `#[ts(concrete)]` an attribute on the type, not the generic parameters

WilsonGramer:
The syntax looks like this now:

```rust
#[derive(TS)]
#[ts(export, concrete(T = TsDriver))]
struct MyStruct<T: Driver> {
    u: T::Info,
}
```


escritorio-gustavo:
> I guess they can also take the stand that they'd like to keep the bug to discourage the syntax, but that feels pretty wrong.

Agreed, serde compatibility is pretty much a requirement for every Rust library. Intentionally keeping a bug that prohibits the use of valid Rust syntax on the basis that they don't like said syntax is very weird.

PR #267, opened 2024-03-13
Add leading colon to `ts_rs` path in macro expansion

escritorio-gustavo:
## Goal

If, for some (very bespoke and cursed) reason, a user creates a `mod` called `ts_rs` and has it in scope, it becomes impossible to
derive `TS`, all the `ts_rs::<something>` paths will look inside that `mod` instead of our library.
This PR aims to make such (cursed) code compile

## Changes

All paths to `ts_rs` generated by `ts_rs_macros` now have two leading colons, e.g.: `::ts_rs::TS`

## Checklist

- [x] I have followed the steps listed in the [Contributing guide](https://github.com/Aleph-Alpha/ts-rs/blob/main/CHANGELOG.md).

### DIFF
--- a/macros/src/deps.rs
+++ b/macros/src/deps.rs
@@ -12,7 +12,7 @@ impl Dependencies {
     /// Adds all dependencies from the given type
     pub fn append_from(&mut self, ty: &Type) {
         self.dependencies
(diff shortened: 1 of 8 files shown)


NyxCode (approved):
LGTM!

PR #273, opened 2024-03-19
feat(impls): add rust_decimal::Decimal impl

### LINKED ISSUE #272: Feature request: rust_decimal::Decimal TS implementation
My project uses [rust_decimal](https://github.com/paupino/rust-decimal) instead of [bigdecimal](https://github.com/akubera/bigdecimal-rs), which seems to be more popular. It might be helpful to provide an implementation for this as well.

Provide an implementation of TS for rust_decimal::Decimal, enabled via "rust-decimal-impl" (or omit the first dash)

**Describe alternatives you've considered**
Downstream packages supply some manual implementation or fallback? 

**Additional context**
I recently began using the crate, so perhaps this addition is unnecessary.

seanpianka:
## Goal

Provide a serde-based implementation of TS for rust_decimal::Decimal
Closes #272 

## Changes

Derive an implementation for rust_decimal::Decimal using feature flag [serde-with-str](https://github.com/paupino/rust-decimal?#serde-with-str).

## Checklist


NyxCode:
I'm a bit confused by `rust-decimal` in general. Their usage of features seems a bit odd to me (though we're abusing them similarly I guess).  

If we want support `rust-decimal` going forward, I think we **need** to seamlessly support all these modes - and I'm not sure that's easily possible.

Re-exposing all features of `rust_decimal`, for example, doesn't seem great.  
`rust_decimal` already has feature gates for a wide variety of libraries, so maybe it'd be worth a shot to open a PR there. In the `rust_decimal` crate, this could be implemented nicely using `#[cfg(feature = "..")]` annotations.

NyxCode:
Closing this for now. I think we established that we can't support this properly. 
Either `rust_decimal` implements `TS` behind a feature-flag, or users will have to use `#[ts(as = ..)]` or `#[ts(type = "..")]`.

PR #274, opened 2024-03-19
Option to rename source of `TS` trait

aumetra:
## Goal

The goal of this PR is to provide an option which allows for re-exports of the `ts_rs::TS` trait.  
Similar to serde's `#[serde(crate = "other_serde")]` attribute.

Motivation is us evaluating `ts-rs` for generating TypeScript types directly without an intermediate JSON Schema step. We generate the derive annotations via an own procedural macro, which is why we need to re-export the crate.

## Changes

This PR introduces the new `crate_rename` attribute to the derive macro.  
The invocation looks similar to this:

```rust
#[derive(TS)]
#[ts(crate_rename = "::my_crate::ts_rs")]
```

## Checklist

(Will add the documentation if this is actually a feature that would be accepted)


aumetra:
This PR is missing like 90% of all renames. So far from complete.  
Will complete and clean it up once it is clear whether this would be accepted.

macros/src/attr/enum.rs
```diff
@@ -1,16 +1,17 @@
...
 use crate::{
     attr::{parse_assign_inflection, parse_assign_str, parse_concrete, Inflection},
     utils::{parse_attrs, parse_docs},
 };
 
-use super::parse_bound;
-
 #[derive(Default)]
 pub struct EnumAttr {
+    pub crate_rename: Option<TokenStream>,
```
escritorio-gustavo, line 14: This should be an `Option<syn::Path>` to allow better parsing and compiler errors when invalid syntax is provided.

macros/src/attr/struct.rs
```diff
@@ -1,16 +1,17 @@
...
 use crate::{
     attr::{parse_assign_str, parse_concrete, Inflection, VariantAttr},
     utils::{parse_attrs, parse_docs},
 };
 
-use super::parse_bound;
-
 #[derive(Default, Clone)]
 pub struct StructAttr {
+    pub crate_rename: Option<TokenStream>,
```
escritorio-gustavo, line 14: Same here, use `Option<syn::Path>` instead

escritorio-gustavo:
Hey, thanks for the PR! I've left a couple of changes I'd like to see here, apart from those, I don't see an issue with adding this feature.

aumetra:
> I've left a couple of changes I'd like to see here, apart from those, I don't see an issue with adding this feature

Great! Then I will check on how to sync the same crate name across all the different code paths and will take care of your requested changes.

NyxCode:
I'm open to adding this as well.  👍

NyxCode:
Interestingly, this would come in handy in #276, allowing to `#[derive(TS)]` within `ts-rs` by using `#[ts(crate = "")]`

escritorio-gustavo (approved):

aumetra:
Oh, you got there before I did. Cool!  
For a test, do you think introducing a separate test crate, such as `ts-test-rename` would be sensible and then importing it as a dependency via

```toml
ts-rename = { package = "ts-rs", path = "../ts-rs" }
```

That way you can be 100% sure the rename code works. The test I made is a bit.. lackluster

escritorio-gustavo:
> Oh, you got there before I did. Cool! For a test, do you think introducing a separate test crate, such as `ts-test-rename` would be sensible and then importing it as a dependency via
> 
> ```toml
> ts-rename = { package = "ts-rs", path = "../ts-rs" }
> ```
> 
> That way you can be 100% sure the rename code works. The test I made is a bit.. lackluster

This is a very good idea! You can do that inside the `e2e` directory

NyxCode:
Great work guys, awesome!  
I've taken the liberty to make `crate_rename` private and replace its use with a getter, `crate_rename() -> Path`. I didn't love that we had these `unwrap()`s everywhere.

NyxCode (approved):

escritorio-gustavo:
> Great work guys, awesome! I've taken the liberty to make `crate_rename` private and replace its use with a getter, `crate_rename() -> Path`. I didn't love that we had these `unwrap()`s everywhere.

Nice! I didn't like having so many unwraps either, glad to see them gone!

PR #276, opened 2024-03-19
Feature: Add support for types from `serde_json`

NyxCode:
## Goal
This PR adds support for types from `serde_json`: `serde_json::Value`, `serde_json::Number` and `serde_json::Map`.  

Most common (and most tricky) of them is `serde_json::Value`.  
Right now, the easiest way to get a struct containing a `Value` to implement `TS` is to slap a `#[ts(type = "any")]` on the fields.  
However, that's not great, since a JSON object is actually not `any`.

If a user wanted to get a more precise type for it, they might have tried this:
```rs
#[derive(TS)]
#[ts(untagged, export_to = "serde_json/")]
enum TsJsonValue {
    String(String),
    Number(i32),
    Array(Vec<TsJsonValue>),
    Object(HashMap<String, TsJsonValue>),
}

#[derive(TS)]
struct SomeStruct {
    #[ts(as = "TsJsonValue")]
    field: serde_json::Value,
}
```

Interestingly, tsc doesn't want to compile `TsJsonValue.ts`. This is one of the (rare) cases where `Record<string, TsJsonValue>`, which is what we generate, doesn't work, but `{ [key: string]: TsJsonValue }` would.

## Changes

In essence, all this PR does is the code above, except that `TS` had to be implemented manually for `serde_json::Value`.

Unlike all previous `..-impl` features, this one actually exposes an exportable type, `JsonValue`.  
Before, all impls were simply and just generated e.g `number`. The one for `serde_json::Value`, however, is recursive, and therefore needs an actual name. 

If a user exports a struct containing `serde_json::Value`, `JsonValue.ts` will get exported to `./serde_json/`.

## Alternatives

We could, instead of merging this now, first explore if generating `{ [key: K]: V }` for maps makes sense.  
Then, this PR would not require the manual TS implementation (If we made `#[derive(TS)]` work inside `ts_rs`, which it currently does not).   
(#134)

## Checklist


ts-rs/src/serde_json.rs
```diff
@@ -0,0 +1,48 @@
...
+use super::{impl_primitives, impl_shadow, typelist::TypeList, TS};
+
+// Manually implement `TS` for `Value`.
+// Defining an untagged enum and deriving `TS` doesn't work for the following reasons:
+// - `#[derive(TS)]` doesn't work in this crate, since the macro generates `::ts_rs::TS`
+// - `Value` is self-referential, and in this case (an untagged enum), tsc doesn't compile
+//   `type JsonValue = ... | Record<string, JsonValue>`, but `{ [x: string]: JsonValue }` works.
+impl TS for serde_json::Value {
+    type WithoutGenerics = Self;
+
```
NyxCode, line 12: The `impl` is pretty exactly like the one generated by 
```rs
#[derive(TS)]
#[ts(untagged, export_to = "serde_json/")]
enum JsonValue {
    String(String),
    Number(i32),
    Array(Vec<TsJsonValue>),
    Object(HashMap<String, JsonValue>),
}
```
with the exception that it uses `{ [key: string]: JsonValue }` instead of `Record<string, JsonValue>`.

NyxCode:
I just realized that what would work instead of the manual implementation is
```rs
#[derive(TS)]
#[ts(untagged, export_to = "serde_json/")]
enum JsonValue {
    String(String),
    Number(i32),
    Array(Vec<JsonValue>),
    Object(#[ts(type = "{ [key: string]: JsonValue }")] ()),
}
```
since the `Array` variant already creates a dependency on `JsonValue`.

escritorio-gustavo (approved):
I really like these changes, if you want I can push some commits to #274 so we can merge it faster as well as create a PR for #134

NyxCode:
I've merged #277 and #274 into this branch to see how that'd look like.  
However, now imports don't always work, and I'm not sure yet why.

NyxCode:
Nevermind - `impl_shadow` did not generate `fn output_path()`. Never needed it before, because all types we shadowed weren't exportable.

escritorio-gustavo:
Awesome! Can this PR be merged or is there anything else you'd still like to do?

ts-rs/src/serde_json.rs
```diff
@@ -0,0 +1,21 @@
+use std::collections::HashMap;
+
+use super::{impl_primitives, impl_shadow, TS};
+
+#[derive(TS)]
+#[ts(
+    crate = "crate",
+    rename = "Value",
```
NyxCode, line 8: Last thing I'm unsure about here: Is `Value` to vague? We currently don't/can't deal with name collisions, so maybe `JsonValue` might be a better name.

escritorio-gustavo: `JsonValue` would be a better name, but given the use of `#[ts(export_to = "serde_json/")]` I think a name collision is really unlikely

NyxCode: Alright, lemme quickly rename it. The name collision i'm more worried about is 
```ts
import type { Value } from "../serde_json/Value";
import type { Value } from "../somewhere/else/Value";
```

escritorio-gustavo: Oh, I see. I though you were worried about the file getting overwritten

NyxCode:
Alright, let's get this merged!  
I'm super happy with how #277 and #274 came together.  
And we're finally getting some use out of the effort we put in to export paths and recusrive exports.

NyxCode:
I guess this demonstrates that libraries can really just slap `#[derive(TS)]` on their types. Maybe, in the future, we won't have to add support for every small crate, but they add support for us :D

escritorio-gustavo:
> I guess this demonstrates that libraries can really just slap `#[derive(TS)]` on their types. Maybe, in the future, we won't have to add support for every small crate, but they add support for us :D

That will be awesome to see!

PR #277, opened 2024-03-20
Replace `Record<K, V>` with `{ [key: K]: V }`

### LINKED ISSUE #134: Use `{ [key: K]: V }` instead or `Record<K, V>` for Maps

escritorio-gustavo:
## Goal

Allow recursive `HashMap` types by changing their inlining from `Record<K, V>` to `{ [key: K]: V }`
Closes #134

## Changes

The `impl` block for `HashMap` was altered

## Checklist


ts-rs/src/lib.rs
```diff
@@ -863,15 +863,15 @@ impl<K: TS, V: TS, H> TS for HashMap<K, V, H> {
     type WithoutGenerics = HashMap<Dummy, Dummy>;
 
     fn ident() -> String {
-        "Record".to_owned()
+        "{ [key: any]: any }".to_owned()
```
escritorio-gustavo, line 866: I actually have no idea what this method should return. Maybe it should `panic!`?

NyxCode:
As far as I can tell, I think this is a minor change (e.g 8.0.1), and shouldn't break anything.

PR #280, opened 2024-03-21
Support serde with

### LINKED ISSUE #275: Can we support `#[serde(with = "..")]`?
Would it make sense for us to parse `#[serde(with = "..")]` as an alias for `#[ts(as = "..")]`?  

`serde` seems to, when encountering a field with `#[serde(with = "X")]`, (de)serialize with `X::(de)serialize`.  
That means that `#[serde(with)]` can point to a module with two functions, or two a struct implementing `Serialize` and `Deserialize`.  

In the 1st case, there's nothing we can do. For case 2, however, we could interpret that as `#[ts(as = "X")]`.

So, to support this, we'd need to figure out if `X` is a type or a module path. We could decide that based on capitalization (`a::b::C` is probably a type, and `a::b::c` is probably a module). 
If we determine that it's a module, we ignore it. If it's a type, we interpret the annotation as `#[ts(as = "..")]`. It's not bullet-proof, but probably good enough?

If we went down that road, it'd make sense to have `#[ts(type = "..")]` or `#[ts(as = "..")]` take prescedence over `#[serde(with = "..")] `, overriding it.

Happy to hear any input on this!

## Links
- [serde docs for `#[serde(with)]`](https://serde.rs/field-attrs.html#with)
- [serde docs for remote types](https://serde.rs/remote-derive.html)

escritorio-gustavo:
## Goal

Ensure the `#[serde(with = "...")]` attribute is properly handled in a way that prevents users from accidentally generating wrong TS type definitions by forgetting to add `#[ts(as = "...")]`/`#[ts(type = "...")]`,

### Why?

The `#[serde(with = "...")]` attribute changes how the type is serialized, therefore, it changes what type will be received by TypeScript, so the type generated by default will most likely be incorrect. The user must use `#[ts(as = "...")]` or `#[ts(type = "...")]` to ensure their type definition matches the type generated by serde

### Why not make `#[serde(with = "...")]` an alias for `#[ts(as = "...")]`

`#[ts(as = "...")]` only accepts types, where as `#[serde(with = "...")]` accepts either a type or a path to a module containing both a `serialize` and a `deserialize` function, so the attributes are not compatible and thus cannot just alias each other

Closes #275

## Changes

When the `serde-compat` feature is enabled, using  `#[serde(with = "...")]` will result in a compiler error if either `#[ts(as = "...")]` or `#[ts(type = "...")]` is not added

## Checklist

## Todo

- [ ] Find a way to test the compiler error case (`compiletest_rs`?)

### DIFF
--- a/macros/src/attr/field.rs
+++ b/macros/src/attr/field.rs
@@ -121,6 +121,9 @@ impl_parse! {
                 parse_assign_str(input)?;
             }
         },
-        "with" => out.0.using_serde_with = true,
+        "with" => {
+            parse_assign_str(input)?;
+            out.0.using_serde_with = true;
+        },
     }
 }
--- a/macros/src/attr/field.rs
+++ b/macros/src/attr/field.rs
@@ -13,6 +13,9 @@ pub struct FieldAttr {
     pub optional: Optional,
     pub flatten: bool,
     pub docs: String,
+
+    #[cfg(feature = "serde-compat")]
+    pub using_serde_with: bool,
 }
 
 /// Indicates whether the field is marked with `#[ts(optional)]`.
@@ -52,6 +55,9 @@ impl FieldAttr {
             optional: Optional { optional, nullable },
             flatten,
             docs,
+
+            #[cfg(feature = "serde-compat")]
+            using_serde_with,
         }: FieldAttr,
     ) {
         self.rename = self.rename.take().or(rename);
@@ -65,6 +71,11 @@ impl FieldAttr {
         };
         self.flatten |= flatten;
         self.docs.push_str(&docs);
+
+        #[cfg(feature = "serde-compat")]
+        {
+            self.using_serde_with = self.using_serde_with || using_serde_with;
+        }
     }
 }
 
@@ -110,5 +121,6 @@ impl_parse! {
                 parse_assign_str(input)?;
             }
         },
+        "with" => out.0.using_serde_with = true,
     }
 }
--- a/macros/src/attr/struct.rs
+++ b/macros/src/attr/struct.rs
@@ -3,9 +3,8 @@ use std::collections::HashMap;
 use syn::{parse_quote, Attribute, Ident, Path, Result, Type, WherePredicate};
 
 use super::{parse_assign_from_str, parse_bound, parse_concrete};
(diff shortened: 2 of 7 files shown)


NyxCode:
Huh, this is much less complicated than I thought it might end up being, nice!  

> #[serde(with = "...")] is a valid variant attribute, but I'm not yet clear on how to implement it, as neither #[ts(as = "...")] or #[ts(type = "...")] are valid variant attributes.

Hm. I guess we *could* at support for them for variants, or leave this edge-case for now. Will have to look into how complicated supporting these attributes for variants would be first, though I think it might be pretty straight forward.

> Find a way to test the compiler error case (Maybe compiletest_rs or trybuild?)

I think that's supported in doctests with "\`\`\`compile_fail  ....`". Not super clean, but it's simple, so maybe that's enough?

escritorio-gustavo:
> I think that's supported in doctests with "```compile_fail ....`". Not super clean, but it's simple, so maybe that's enough?

It is, but I'm not sure if a doctest in the `tests` directory would be executed by `cargo test`

escritorio-gustavo:
> Hm. I guess we _could_ at support for them for variants, or leave this edge-case for now. Will have to look into how complicated supporting these attributes for variants would be first, though I think it might be pretty straight forward.

That might require two new PRs, one for `#[ts(as = "...")]`/`#[ts(type = "...")]` and another for `#[serde(with = "...")]`. Otherwise I think this one would get too large

NyxCode:
Agreed!

NyxCode (approved):
Didn't find the time to take a closer look earlier, sorry about that.  
Everything looks good to me here! I was still wishing that we *could* actually support `#[serde(with = "..")]`, but this is still better than the status quo and is actually robust. I'm sold!

LZQCN:
Good PR, Please release a new version that includes these updates. I would really like to use them in my project. Thank you!

escritorio-gustavo:
Hello @LZQCN! These changes will eventually be released with version 9 of the library, but we still need to complete #284, which covers this same feature on enum variants. It hasn't been merged yet because I couldn't quite figure out a way to test it and compare the results with `serde`'s output.

PR #281, opened 2024-03-22
Fix generic impl

aumetra:
## Goal

When deriving `ts_rs::TS` for a struct with a generic type parameter, the code expanded to requiring the `ts_rs::TS` trait to be in scope.  
This should be fixed.

## Changes

This PR fixes this by using the newly introduced `crate_rename` path, interpolating the path to be `#crate_rename::TS`

## Checklist


escritorio-gustavo (approved):
LGTM! @NyxCode, I think we should make a 8.1.0 release after merging this PR. What do you think?

NyxCode:
whoops, guess we missed that.  

> I think we should make a 8.1.0 release after merging this PR. What do you think?

I'd be open to. Let me first take a look at #282 and #280, to see if we can include them as well.

escritorio-gustavo:
> @escritorio-gustavo I think #280 needs a bit more time, right? Should I publish a 8.1.0 from the current master as-is?

Agreed, I'd love to have that fully functioning before it gets to a release

PR #282, opened 2024-03-22
Fix bug in enum flattening

escritorio-gustavo:
## Goal

When a struct containing a flattened enum was itself flattened, the type generated would be missing the enum
That is, the following code

```rs
#[derive(TS)]
struct FooExternally {
    qux: i32,
    #[ts(flatten)]
    baz: BarExternally,
    biz: Option<String>,
}

#[derive(TS)]
enum BarExternally {
    Baz { a: i32, a2: String },
    Biz { b: bool },
    Buz { c: String, d: Option<i32> },
}

#[derive(TS)]
#[ts(export, export_to = "enum_flattening/externally_tagged/")]
struct NestedExternally {
    #[ts(flatten)]
    a: FooExternally,
    u: u32,
}
```

Would generate the following TS:
```ts
export type NestedExternally = { u: number, qux: number, biz: string | null, };
```

When it should generate
```ts
export type NestedExternally = { u: number, qux: number, biz: string | null, } & (
    | { "Baz": { a: number, a2: string, } }
    | { "Biz": { b: boolean, } }
    | { "Buz": { c: string, d: number | null, } }
);
```

## Changes

Updated the generation of `inline_flattened` in `macros/src/types/named.rs` to be similar to `inline`, except for not removing parenthesis around lonely flattened fields, as that caused issues when flattening a struct that contained only a flattened enum. (The `replace(" } & { ", " ")` would merge the fields with the first enum variant)

## Checklist

### DIFF
--- a/ts-rs/tests/enum_flattening_nested.rs
+++ b/ts-rs/tests/enum_flattening_nested.rs
@@ -103,15 +103,30 @@ struct NestedAdjecently {
     u: u32,
 }
 
+#[derive(TS)]
+#[cfg_attr(feature = "serde-compat", derive(Serialize))]
+#[ts(export, export_to = "enum_flattening_nested/externally_tagged/")]
+struct NestedAdjecentlyLonely {
+    #[cfg_attr(feature = "serde-compat", serde(flatten))]
+    #[cfg_attr(not(feature = "serde-compat"), ts(flatten))]
+    a: FooAdjecently,
+}
+
 #[test]
 fn adjacently_tagged() {
     assert_eq!(
         FooAdjecently::inline(),
         r#"{ "type": "Baz", "stuff": { a: number, a2: string, } } | { "type": "Biz", "stuff": { b: boolean, } } | { c: string, d: number | null, }"#
     );
+
     assert_eq!(
         NestedAdjecently::inline(),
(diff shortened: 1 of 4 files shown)


ts-rs/tests/enum_flattening_nested.rs
```diff
@@ -0,0 +1,198 @@
...
+use ts_rs::TS;
+
+#[derive(TS)]
+#[cfg_attr(feature = "serde-compat", derive(Serialize))]
+#[ts(export, export_to = "enum_flattening_nested/externally_tagged/")]
+struct FooExternally {
+    #[cfg_attr(feature = "serde-compat", serde(flatten))]
+    #[cfg_attr(not(feature = "serde-compat"), ts(flatten))]
+    baz: BarExternally,
+}
```
escritorio-gustavo, line 14: This test file checks what happens when the flattened enum is the only field in the flattened struct

NyxCode:
Nice, thanks!  
I never realized that `inline_flattened` would just drop flattened fields. Amazing that the test suite didnt catch that.

NyxCode (approved):

escritorio-gustavo:
> Nice, thanks! I never realized that `inline_flattened` would just drop flattened fields. Amazing that the test suite didnt catch that.

Me neither! I only realized because I was looking at some old PRs and when I got to #206 I went "wtf was I thinking" lol

PR #284, opened 2024-03-22
Support `#[ts(as = "...")]` and `#[ts(type = "...")]` on enum variants

escritorio-gustavo:
## Goal

Add support for using  `#[ts(as = "...")]` and `#[ts(type = "...")]` on enum variants to make support for `#[serde(with = "...")]` (#280) possible here too

## Changes

`#[ts(as = "...")]` and `#[ts(type = "...")]` are now parsed and processed in enum variants.

Using them currently respects the tag enum representation. That is, neither `#[ts(as = "...")]` nor `#[ts(type = "...")]` will eliminate the "tag" or "content" properties the variant would have without them, as that can be achieved by using `#[ts(untagged)]` on the variant. I don't yet know if this is how serde works, so this can be changed to maintain serde compatibility

## Checklist


escritorio-gustavo:
I may need some help on the tests for this. I can't even get `#[serde(with = "...")]` on variants to compile atm

NyxCode:
would love to take a loot at this once i find some time 👍

escritorio-gustavo:
@NyxCode, I think I figured out how `#{serde(with = ...)]` is used in variants. Looking at the last commit, do you think the current behavior of `#[ts(as = ...)]` is correct? I think it is, but I still don't quite understand `#{serde(with = ...)]` deeply enough to be sure

JonasFocke01:
I tested my usecase from #352 and can confirm that this works as expected

gustavo-shigueo:
Hi @JonasFocke01, I've just fixed a merge conflict and caused a couple of compiler errors in the process, can you check if it still works?

JonasFocke01:
@gustavo-shigueo Workds still as expected :+1:

NyxCode (approved):
nicely done!

PR #286, opened 2024-03-27
add support for top level type overrides

dr-bonez:
## Goal

This adds support for easily defining a struct or enum's type definition manually at the top level using the `type = ".."` attribute. This is useful for when you have a custom serialize/deserialize and don't want to manually implement the `TS` trait.

## Changes

Added `type_override` to the container attributes, and defined an early-returned implementation for if this attribute is specified

## Checklist

### DIFF
--- a/macros/src/types/enum.rs
+++ b/macros/src/types/enum.rs
@@ -5,7 +5,8 @@ use syn::{Fields, ItemEnum, Variant};
 use crate::{
     attr::{EnumAttr, FieldAttr, StructAttr, Tagged, VariantAttr},
     deps::Dependencies,
-    types, DerivedTS,
+    types::{self, type_override},
+    DerivedTS,
 };
 
 pub(crate) fn r#enum_def(s: &ItemEnum) -> syn::Result<DerivedTS> {
@@ -17,6 +18,10 @@ pub(crate) fn r#enum_def(s: &ItemEnum) -> syn::Result<DerivedTS> {
         Some(existing) => existing.clone(),
         None => s.ident.to_string(),
     };
+
+    if let Some(t_o) = &enum_attr.type_override {
+        return type_override::type_override_enum(&enum_attr, &name, t_o);
+    }
 
     if s.variants.is_empty() {
         return Ok(empty_enum(name, enum_attr));
--- a/macros/src/types/enum.rs
+++ b/macros/src/types/enum.rs
@@ -19,8 +19,8 @@ pub(crate) fn r#enum_def(s: &ItemEnum) -> syn::Result<DerivedTS> {
         None => s.ident.to_string(),
     };
 
-    if let Some(t_o) = &enum_attr.type_override {
-        return type_override::type_override_enum(&enum_attr, &name, t_o);
+    if let Some(attr_type_override) = &enum_attr.type_override {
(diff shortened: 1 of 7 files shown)


NyxCode:
I like this! It's a simple change and it's cleanly implemented!  
What do you think @escritorio-gustavo?

macros/src/types/mod.rs
```diff
@@ -18,6 +19,10 @@ pub(crate) fn struct_def(s: &ItemStruct) -> Result<DerivedTS> {
 
 fn type_def(attr: &StructAttr, ident: &Ident, fields: &Fields) -> Result<DerivedTS> {
     let name = attr.rename.clone().unwrap_or_else(|| to_ts_ident(ident));
+    if let Some(t_o) = &attr.type_override {
```
escritorio-gustavo, line 22: I think it's better to call the variable `type_override` instead of `t_o`. These abbreviatons are usually bad and they tend to become a habit if allowed

macros/src/types/type_override.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+pub(crate) fn type_override_enum(
+    attr: &EnumAttr,
+    name: &str,
+    type_override: &str,
+) -> Result<DerivedTS> {
+    if attr.rename_all.is_some() {
+        syn_err!("`rename_all` is not compatible with `type`");
+    }
```
escritorio-gustavo, line 45: `rename_all_fields` should probably be imcompatible as well

escritorio-gustavo (approved):
I have two small nitpicks, otherwise, LGTM

macros/src/types/type_override.rs
escritorio-gustavo: Not only that, but also `tag`, `content` and `untagged`, as well as `tag` for structs

macros/src/types/mod.rs
dr-bonez: this shadows the module name, which I try to avoid. I just couldn't think of a better name for it 😅 
I'm open to suggestions

escritorio-gustavo: Maybe just `ty_override` then

macros/src/types/enum.rs
```diff
@@ -19,8 +19,8 @@ pub(crate) fn r#enum_def(s: &ItemEnum) -> syn::Result<DerivedTS> {
         None => s.ident.to_string(),
     };
 
-    if let Some(t_o) = &enum_attr.type_override {
-        return type_override::type_override_enum(&enum_attr, &name, t_o);
+    if let Some(attr_type_override) = &enum_attr.type_override {
+        return type_override::type_override_enum(&enum_attr, &name, attr_type_override);
```
escritorio-gustavo, line 23: Perfect!

macros/src/types/mod.rs
dr-bonez: I went with `attr_type_override` because it plays better with rust analyzer's parameter name hints

escritorio-gustavo:
@NyxCode is there anything else you'd like to change?

NyxCode (approved):
Nothing from my side!

Great work @dr-bonez, thanks!

PR #287, opened 2024-03-28
Fix `#[ts(rename_all_fields = "...")]` on enums containing tuple or unit variants

escritorio-gustavo:
## Goal

Using `#[ts(rename_all_fields = "...")]` on an enum containing tuple or unit variants would cause the error "`rename_all` is not applicable to newtype/tuple/unit structs" 

## Changes

Only add `rename_all` from `rename_all_fields` if `variant.fields` is `Fields::Named(_)`

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,12 @@
 # master
+
 ### Breaking
+
 ### Features
+
 ### Fixes
 
+Fix `#[ts(rename_all_fields = "...")]` on enums containing tuple or unit variants ([#287](https://github.com/Aleph-Alpha/ts-rs/pull/287))
 
 # 8.1.0
 
--- a/macros/src/attr/variant.rs
+++ b/macros/src/attr/variant.rs
@@ -21,14 +21,8 @@ pub struct SerdeVariantAttr(VariantAttr);
 
 impl VariantAttr {
     pub fn new(attrs: &[Attribute], enum_attr: &EnumAttr) -> Result<Self> {
-        let mut result = Self::default();
-        parse_attrs(attrs)?.for_each(|a| result.merge(a));
+        let mut result = Self::from_attrs(attrs)?;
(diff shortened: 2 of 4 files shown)


NyxCode:
To summarize my understanding:  
The problem is that when creating `VariantAttr`s in `VariantAttr::new`, we use the `rename_all_fields` of the enum as value for `rename_all`. 
These `VariantAttr`s are then converted into `StructAttr`s, and when they reach `newtype`, we get the error.

There are a couple of thinks that stood out to me.  
This basically duplicates `VariantAttr::new`, though we could fix that by re-writing `VariantAttr::new` to use `VariantAttr::from_attrs` internally.   
Also, the `match`, deciding between `VariantAttr::new` and `VariantAttr::from_attrs` really doesn't feel great.

An alternative way of getting rid of the error would be to remove the check from `newtype`. It probably wouldn't be a big deal to just ignore it if it's there, to be honest. 

If we wanted to keep the error, we could alternatively move these sanity checks, which we have in a lot of places, into the attribute parsing logic. But again, we'd then need to know what we're parsing (a unit struct?). So that'd probably be a larger refactor, and I'm not sure if it'd end up being nice.

So yeah, I'm not super satisfied with this solution, but I kinda don't see a better way of doing it right now.  
If we're going with this, I think we should definetely add a comment to the `match`, and re-write `VariantAttr::new` to use `VariantAttr::from_attrs` to not have that logic duplicated.

escritorio-gustavo:
> This basically duplicates `VariantAttr::new`, though we could fix that by re-writing `VariantAttr::new` to use `VariantAttr::from_attrs` internally.

Agreed! I'll push a commit to do that

NyxCode (approved):
Alright, nice!  
I'd be happy with merging this as-is. Maybe we'll find some nice way to clean all the "... is not compatible with ..." errors up at some point in the future.

PR #290, opened 2024-04-02
Rework attribute parsing

escritorio-gustavo:
## Goal

This PR aims to change the way we parse attributes in a few key ways:
- Introducing a couple of traits for the methods all attribute types share
- Change the `merge` method to remove the mutable reference
  - The reasoning for this is mostly that I think a `for_each` that mutates external data is somewhat confusing, so I changed it to use `fold` (think JS `Array.protoype.reduce`) instead.
- Bring all the attribute compatibility validation to the same file where parsing is done

## Checklist

  - This is a refactor, so there isn't really new anything I can test... We could make tests for whether or not our compiler errors are correct, but we still have to figure out how to do that

### DIFF
--- a/macros/src/attr/struct.rs
+++ b/macros/src/attr/struct.rs
@@ -96,6 +96,16 @@ impl Attr for StructAttr {
             }
         }
 
+        if self.type_as.is_some() {
+            if self.tag.is_some() {
+                syn_err!("`tag` is not compatible with `as`");
+            }
+
+            if self.rename_all.is_some() {
+                syn_err!("`rename_all` is not compatible with `as`");
+            }
+        }
+
         if !matches!(item, Fields::Named(_)) {
             if self.tag.is_some() {
                 syn_err!("`tag` cannot be used with unit or tuple structs");
--- a/macros/src/attr/struct.rs
+++ b/macros/src/attr/struct.rs
@@ -32,7 +32,8 @@ impl StructAttr {
 
         #[cfg(feature = "serde-compat")]
         {
-            result = crate::utils::parse_serde_attrs::<StructAttr>(attrs, result);
+            let serde_attr = crate::utils::parse_serde_attrs::<StructAttr>(attrs);
+            result.merge_with_serde(serde_attr)
         }
         Ok(result)
     }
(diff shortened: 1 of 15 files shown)


macros/src/attr/struct.rs
```diff
@@ -28,67 +28,88 @@ pub struct SerdeStructAttr(StructAttr);
...
+        variant_fields: &Fields,
+    ) -> Self {
         Self {
             crate_rename: Some(enum_attr.crate_rename()),
             rename: variant_attr.rename.clone(),
-            rename_all: variant_attr.rename_all,
+            rename_all: variant_attr.rename_all.or(match variant_fields {
+                Fields::Named(_) => enum_attr.rename_all_fields,
+                Fields::Unnamed(_) | Fields::Unit => None,
+            }),
```
escritorio-gustavo, line 53: I managed to move the `rename_all_fields` match to a better location and get rid of that duplicated mehtod

NyxCode:
I'll take a closer look once I get home, but I love what I've seen so far. Great work!

NyxCode:
Great stuff, nothing here I don't like!  
As far as I'm concerned, you can merge this as-is - unless there's anything else you'd like to do.

escritorio-gustavo:
> Great stuff, nothing here I don't like!

Right as you say that I push a compiler error lmao

NyxCode:
What was the reasoning behind the last commit? Was there a bug I missed?

escritorio-gustavo:
> What was the reasoning behind the last commit? Was there a bug I missed?

I thought it was weird passing the `ts` attribute into `parse_serde_attrs` for the initial value of `fold`, so i changed it to avoid doing that

escritorio-gustavo:
> I think that was the last thing that I wanted to change, as long as you like the latest commit, I think this is ready to merge

Actually, I just realized that the `merge_with_serde` method will cause a pretty annoying difference, which is that to make a serde attr compatible with a ts attr `merge_with_serde` needs to be updated, instead of just the `impl_parse` block

NyxCode:
Yeah, that's true. I too thought passing in the "previous" attrs was a tiny bit akward, but I think I preferred it over having the separate `merge_with_serde` fns.

escritorio-gustavo:
In [ba04528](https://github.com/Aleph-Alpha/ts-rs/pull/290/commits/ba04528662ca13d300b674f5875e1340a672a689) removing `merge_with_serde` and using `merge` broke the merging of `docs`, so I just moved the `parse_docs` part to be after the serde attribute parsing

escritorio-gustavo:
Alright, I think I'm happy with this now!

NyxCode (approved):
So am I!  
Great work, super happy you tackled this!

Issue #289, opened 2024-04-02
TypeList recursion limit


kamadorueda:
**Describe the bug**

Since the upgrade from 7.1.1 to 8.1.0, I'm getting recursion limit errors

```
error[E0275]: overflow evaluating the requirement `impl TypeList well-formed`
```

or

```
reached the recursion limit while instantiating `<(((((..., ...), ...), ...), ...), ...) as TypeList>::for_each::<...>`
```

(on a crate that depends on a crate with `#![recursion_limit = "..."]`)

**To Reproduce**

Derive TS on an enum with 130+ fields, for example:

```rs
#[derive(Debug, ts_rs::TS)]
#[ts(export)]
pub enum Iso4217CurrencyCode {
    AED,
    AFN,
    ALL,
    AMD,
    ANG,
    AOA,
    ARS,
    AUD,
    AWG,
    AZN,
    BAM,
    BBD,
    BDT,
    BGN,
    BHD,
    BIF,
    BMD,
    BND,
    BOB,
    BRL,
    BSD,
    BTN,
    BWP,
    BYN,
    BZD,
    CAD,
    CDF,
    CHF,
    CLP,
    CNY,
    COP,
    CRC,
    CUC,
    CUP,
    CVE,
    CZK,
    DJF,
    DKK,
    DOP,
    DZD,
    EGP,
    ERN,
    ETB,
    EUR,
    FJD,
    FKP,
    GBP,
    GEL,
    GGP,
    GHS,
    GIP,
    GMD,
    GNF,
    GTQ,
    GYD,
    HKD,
    HNL,
    HRK,
    HTG,
    HUF,
    IDR,
    ILS,
    IMP,
    INR,
    IQD,
    IRR,
    ISK,
    JEP,
    JMD,
    JOD,
    JPY,
    KES,
    KGS,
    KHR,
    KMF,
    KPW,
    KRW,
    KWD,
    KYD,
    KZT,
    LAK,
    LBP,
    LKR,
    LRD,
    LSL,
    LYD,
    MAD,
    MDL,
    MGA,
    MKD,
    MMK,
    MNT,
    MOP,
    MRU,
    MUR,
    MVR,
    MWK,
    MXN,
    MYR,
    MZN,
    NAD,
    NGN,
    NIO,
    NOK,
    NPR,
    NZD,
    OMR,
    PAB,
    PEN,
    PGK,
    PHP,
    PKR,
    PLN,
    PYG,
    QAR,
    RON,
[...]

PR #292, opened 2024-04-02
Generate somewhat cleaner `TypeList`s 

NyxCode:
See #291, #289

## Goal

Generate `TypeList`s containing less duplicates. This speeds up generation of imports, and avoids rustc recursion limits.

This does not aim to fix the "recursion limit" problem completely. Before, ~128 types caused the issue, while now, it'd only be caused by ~128 **unique** types.

## Changes

- Re-implement `Dependencies` to manually track "direct", "transitive" and "generics" dependencies, only generating rust code at the very end.

## TODOs
- [ ] Cleanup implementation
  - The current impl is kinda messy. I used `Rc`s to avoid excessive cloning, but a better data structure would probably help here
- [ ] Explore if generating a "perfect" `TypeList` is a cleaner approach
  - I think this can be implemented on top of most of these changes though, only needing to alter the `ToTokens` impl 

## Checklist


NyxCode:
closing in favour of #293

PR #293, opened 2024-04-02
dedup dependencies

kamadorueda:
## Goal

Generate TypeLists containing less duplicates.

Mentions #291 #289 

## Changes

This is an Internal API change, use a HashSet instead of a Vec, and delay formatting until just when it's needed.

## Checklist


NyxCode:
Nice! Pretty much the same as #292, but I think I like this one more.

NyxCode:
I explored if generating a custom `TypeList` impl (which would eliminate the recusion problem completely) would a sensible alternative.  

It's definetely possible, but I think it'd require a breaking change to be nice. What I had in mind was
```rs
fn dependency_types() -> impl TypeList {
  #[derive(Copy, Clone)]
  struct Deps;
  impl TypeList for Deps {
    fn for_each(self, v: &mut impl TypeVisitor) {
      v.visit::<TypeOfField1>();
      v.visit::<TypeOfField2>();
      Something::dependency_types().for_each(v);
      SomethingElse::generics().for_each(v);
    }
  }
  Deps
}
```

This breaks with generic parameters though. The naive `for_each` impl would use them, but it can't. to make that work, `Deps` would need to be generic as well. Definetely something we could do right now, but it's kinda messy.

If we changed `fn dependency_types() -> impl TypeList` to `fn dependency_types(v: &mut impl TypeVisitor)`, these problems should go away. That would remove one level of indirection.

So tl;dr: I think we should go ahead with this, and keep this change in mind if someone still runs into the issue or for a breaking 9.0 release some time in the future.

escritorio-gustavo (approved):

NyxCode:
Note to myself: Add two tests from other branch.

ts-rs/tests/recursion_limit.rs
```diff
@@ -1,170 +1,82 @@
...
+
+use ts_rs::{
+    typelist::{TypeList, TypeVisitor},
+    TS,
+};
+
+#[rustfmt::skip]
 #[derive(Debug, ts_rs::TS)]
-#[ts(export)]
+#[ts(export, export_to = "very_big_types")]
```
escritorio-gustavo, line 10: Missing trailing slash here

NyxCode: :D you were right that this is a big footgun

NyxCode:
So, I've added some reference counting (probably doesn't have a big performance impact, but it felt like the right thing to do), added the test from the other branch and used a `HashSet` for `types` as well.  

From my side, this is ready to merge.

ts-rs/tests/recursion_limit.rs
escritorio-gustavo: Sidenote, I didn't realize this would break the format feature. I guess this confirms they use the path only to check the extension for `.ts`

NyxCode:
Thanks @kamadorueda for getting this of the ground!

PR #298, opened 2024-04-05
Fix inconsistencies with `serde` and add `SCREAMING-KEBAB-CASE`

### LINKED ISSUE #297: bug: Using `#[ts(rename_all = "...")]` `snake_case`, `SCREAMING_SNAKE_CASE` and `kebab-case` is not compatible with serde
**Describe the bug**
A few months ago, issue #165 reported inconsistencies in the implementation of `inflector`'s `camelCase` in comparison with `serde`'s. It turns out its `snake_case` implementation is also not consistent, which affects `snake_case`, `SCREAMING_SNAKE_CASE` (a.k.a. `snake_case.to_ascii_uppercase()`) and `kebab-case` (a.k.a. `snake_case.replace('_', "-")`). Also, we're missing `SCREAMING-KEBAB-CASE`

**To Reproduce**
```rs
#[derive(TS, Default, serde::Serialize)]
#[ts(export, export_to = "struct_rename/", rename_all = "kebab-case")]
#[serde(rename_all = "kebab-case")]
struct RenameAllScreamingKebab {
    crc32c_hash: i32,
    some_field: i32,
    some_other_field: i32,
}
```

**Expected behavior**
```ts
export type RenameAllScreamingKebab = {
    "crc32c-hash": number,
    "some-field": number,
    "some-other-field": number,
};
```

**Actual behavior**
```ts
export type RenameAllScreamingKebab = {
    "crc-3-2c-hash": number,
    "some-field": number,
    "some-other-field": number,
};
```

escritorio-gustavo:
## Goal

This PR aims to fix a few inconsistencies with `serde` caused by our use of `inflector`:
- Fix `snake_case` inconsistency
- Fix `SCREAMING_SNAKE_CASE` inconsistency
- Fix `kebab-case` inconsistency
- Add `SCREMING-KEBAB-CASE`
- Remove dependency on `inflector`
- `serde` only accepts the `rename_all` rule if the case is written exactly as described (i.e. `CAMELCASE`, `Kebab-case`, etc. are not allowed), where as `ts` accepts any changes in case, dashes and underscores. This was changed for consistency with `serde`

Closes #297

## Changes

- Manually implement `snake_case` like in `serde`
- Change `parse_assign_inflection` to only accept cases written strictly like in the docs
- Add `Inflection::ScreamingKebab`
- Remove dependency on `inflector`

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,14 @@
 
 ### Breaking
 
+- Fix incompatibility with serde for `snake_case`, `kebab-case` and `SCREAMING_SNAKE_CASE` ([#298](https://github.com/Aleph-Alpha/ts-rs/pull/298))
+- `#[ts(rename_all = "...")]` no longer accepts variations in the string's casing, dashes and underscores to make behavior consistent with serde ([#298](https://github.com/Aleph-Alpha/ts-rs/pull/298))
+
 ### Features
 
 - Add support for `#[ts(type = "..")]` directly on structs and enums ([#286](https://github.com/Aleph-Alpha/ts-rs/pull/286))
 - Add support for `#[ts(as = "..")]` directly on structs and enums ([#288](https://github.com/Aleph-Alpha/ts-rs/pull/288))
+- Add support for `#[ts(rename_all = "SCREAMING-KEBAB-CASE")]` ([#298](https://github.com/Aleph-Alpha/ts-rs/pull/298))
 
 ### Fixes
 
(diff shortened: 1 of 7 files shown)


escritorio-gustavo:
>  `serde` only accepts the `rename_all` rule if the case is written exactly as described (i.e. `CAMELCASE`, `Kebab-case`, etc. are not allowed), where as `ts` accepts any changes in case, dashes and underscores. This was changed for consistency with `serde`

@NyxCode this is the only "unnecessary" change here, though I think it's beneficial, as it forces the user to use the case they are setting `rename_all` to, which helps make sure they are using the correct one (e.g.: a common mistake is using `CamelCase` instead of `PascalCase`)

NyxCode:
Nice change, thanks!  
I'm not sure regarding the breaking change, but I think it's alright, since we publish major versions relatively often anyway.

NyxCode (approved):
great stuff!

Issue #175, opened 2023-10-19
Optional fields with `#[serde(default)]` for deserialization


BenJeau:
Hi thanks for the library!

For a struct with serde default for some fields, the fields becomes optional when Deserializing and if not define will use the default value. I think the resulting TypeScript code should reflect that. The following Rust code:

```rust
#[derive(Deserialize, TS)]
#[ts(export)]
pub struct Request {
    url: String,
    #[serde(default)]
    method: Method,
    #[serde(default = "default_timeout")]
    timeout: u32,
}

#[derive(Deserialize, TS, Default)]
#[ts(export)]
enum Method {
    #[default]
    Get,
    Post,
    Put,
    Delete,
}

fn default_timeout() -> u32  {
    10_000
}
```

Should result in the following TypeScript types if we only care about deserialization

```ts
interface Request {
    url: string;
    method?: Method;
    timeout?: number;
}
```

escritorio-gustavo:
Use `#[ts(optional)]`

abhay-agarwal:
For individual fields labeled as `#[serde(default)]`, ts_rs could label the fields as optional. Currently a `#[ts(optional)]` marker only works on `Option<>` types, though. This would help me specifically with boolean types which are often given as optional types and only serialized if `true`.

escritorio-gustavo:
> For individual fields labeled as `#[serde(default)]`, ts_rs could label the fields as optional. Currently a `#[ts(optional)]` marker only works on `Option<>` types, though. This would help me specifically with boolean types which are often given as optional types and only serialized if `true`.

You could do it like this:

```rs
use serde::Serialize;

#[derive(TS, Serialize)]
#[ts(export)]
struct Foo {
    #[ts(optional)]
    #[serde(skip_serializing_if = "should_skip", default = "Default::default")]
    my_optional_bool: Option<bool>,
}

fn should_skip(field: &Option<bool>) -> bool {
    matches!(field, None | Some(false))
}
```

abhay-agarwal:
Right, this is what we were doing before. However, because Rust doesn't have a `True` object, our code was required to "know" that the Option was always Option(true) or None, but never Option(false), and sometimes we would unwittingly add Option(false) in there if, for example, option chaining wasn't correctly written. So my preference would be to allow ts_rs to generate the question mark property if there was a serde(default), as that should in theory allow for the property to be declared optional.

escritorio-gustavo:
[...]

PR #299, opened 2024-04-08
Infered types (`_`) in `#[ts(as = "...")]`

escritorio-gustavo:
## Goal

This PR aims to allow the use of infered types (`_`) inside the `#[ts(as = "...")]` attribute. The `_` type will be interpreted as the field's original type. E.g.:

```rs
#[derive(TS)]
struct Foo {
    #[ts(as = "Option<_>"]
    bar: i32,
}
```

is the same as:
```rs
#[derive(TS)]
struct Foo {
    #[ts(as = "Option<i32>"]
    bar: i32,
}
```
but without having to repeat the original type.

This allows us to somewhat support using `#[ts(optional)]` with any type, by using `#[ts(as = "Option<_>", optional)]` without requiring us to change the semantics of `#[ts(optional)]`

## Changes

Added a recursive function to traverse the provided type and replace `Type::Infer(_)` with the field's original type

## Checklist


NyxCode:
Interesting! So `_` refers to the original type, right?

NyxCode:
I do think this is pretty neat, though I think it might be a bit unintuitive for anyone who doesn't know what `_` does in that context.  
Have you considered a different syntax, e.g `#[ts(as = "Option<type>")]`. I feel like we might be able to come up with something more intuitive, but I'm not sure if the extra effort/complexity involved in parsing this would be worth it.

gustavo-shigueo:
> Interesting! So `_` refers to the original type, right?

Yes, this is mostly a way to resolve the discussion in #175

abhay-agarwal:
I think the understore is super intuitive, as it's already widely used throughout rust syntax. So the code is instantly readable. The other nice thing is that you could, for example, turn a type into a vector, such as `as = Vec<_>`, or `as = Result<_,Something>`, etc.

escritorio-gustavo:
There is also precedent for this behavior in `serde_with` https://github.com/jonasbb/serde_with?tab=readme-ov-file#examples

NyxCode:
Oh, alright! Maybe it's just me then.  
I feel like the semantics of `_` are very different here, but I think you two convinced me!

macros/src/types/mod.rs
```diff
@@ -46,3 +51,122 @@ fn type_def(attr: &StructAttr, ident: &Ident, fields: &Fields) -> Result<Derived
         Fields::Unit => unit::null(attr, &name),
     }
 }
+
+pub(super) fn type_as_infer(type_as: &Type, original_type: &Type) -> Type {
```
NyxCode, line 55: This is perfectly fine, but just an idea: Would changing the signature to `type_as_infer(type_as: &Type, original: &mut Type) -> ()` make the impl cleaner? I think that might get rid of the `parse_quote!` calls.

macros/src/types/newtype.rs
```diff
@@ -27,7 +28,10 @@ pub(crate) fn newtype(attr: &StructAttr, name: &str, fields: &FieldsUnnamed) ->
         return super::unit::null(attr, name);
     }
 
-    let inner_ty = type_as.as_ref().unwrap_or(&inner.ty).clone();
+    let inner_ty = type_as
+        .as_ref()
+        .map(|ty_as| type_as_infer(ty_as, &inner.ty))
+        .unwrap_or_else(|| inner.ty.clone());
```
NyxCode, line 34: Would it be possible to move this into a getter on `FieldAttr`, making the field `type_as` private?

escritorio-gustavo: There's also the `format_variant` function that needs access to `type_as`

macros/src/types/mod.rs
NyxCode: I've put something together in #305 to highlight what I had in mind. I've put the `&mut` on the wrong argument in the snippet above, i think that caused some confusion. Sorry about that.

escritorio-gustavo: > I've put something together in https://github.com/Aleph-Alpha/ts-rs/pull/305 to highlight what I had in mind.

Oh, that makes a lot more sense! I'll merge that in real quick

macros/src/types/enum.rs
```diff
@@ -135,30 +135,16 @@ fn format_variant(
...
-                        (Some(type_override), None) => quote! { #type_override },
-                        (None, Some(type_as)) => {
-                            quote!(<#type_as as #crate_rename::TS>::name())
-                        }
-                        (None, None) => {
-                            let ty = &unnamed.unnamed[0].ty;
+                    let ty = match field_attr.type_override {
+                        Some(type_override) => quote!(#type_override),
+                        None => {
+                            let ty = field_attr.type_as(&field.ty);
```
NyxCode, line 144: I love how this got a lot simpler. Awesome.

NyxCode (approved):
Awesome stuff, great work!

NyxCode:
Also, you were spot on with your take in #175. This solution keeps `#[ts(optional)]` as it was. Specifically, `#[ts(optional)]` still doesn't alter the type, just how it's represented. This seems super clean to me!

NyxCode:
Interestingly, this kinda makes `#[ts(optional = nullable)]` obsolete. It'd be equivalent to 
```rs
#[ts(as = "Option<_>", optional)]
field: Option<i32>,
```

Issue #94, opened 2022-04-11
large integer types should also be typed as numbers


cdmistman:
currently, `i64`, `u64`, `i128`, and `u128` are defined to export to `bigint`, such that:
```rs
#[derive(TS)]
pub struct Foo {
  foo: u64,
}
```
creates a TS declaration like:
```ts
export interface Foo { foo: bigint }
```

However, `foo` may be a `number` instead of a `bigint` eg if using `JSON.parse`, so the interface should actually be:
```ts
export interface Foo { foo: bigint | number }
```

zachmatson:
Worth noting that wasm-bindgen converts u64 arguments/returns to BigInt due to the limitations of JS numbers

HalfVoxel:
> However, foo may be a number instead of a bigint eg if using JSON.parse, so the interface should actually be:

It's worth noting that if you have a i64/u64/etc. and deserialize it to a number in javascript you may get subtle errors as a `number` cannot represent everything in a 64 bit integer. I think bigint is the more correct option here.

cdmistman:
I understand that 64+-bit numbers are usually converted to bigint when sent to JS across the WASM boundary, but when serializing using eg `serde_json`, [the value is not specified as a bigint](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a6401f0dc78b7bd930c67d0d780a635f) (the number isn't written as `5n`, it's written as `5` so `JSON.parse` selects `number` for the deserialized value). 

So when using `ts-rs` for eg an HTTP API, the generated typescript bindings are wrong - it assumes that the value is *always* a bigint, when in fact it can be a bigint *or a number*. Since JS doesn't automatically coerce between the 2 types and they have different properties from each other (eg [one is safe to use in cryptography while the other isn't](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#cryptography)), it should be up to the consumer of the API to decide whether the value should be treated as a bigint or a number.

HalfVoxel:
Sending an u64 over json to javascript is not valid in general. Take a look at this for example:
![image](https://user-images.githubusercontent.com/1144597/178114462-eee05b1f-38ef-46e7-a181-bbfc0d82b665.png)

Therefore, I would say that the only good use case for sending an u64 to javascript is via a path that makes it a bigint, not a number.

[...]

PR #304, opened 2024-04-09
CLI initial concept

### LINKED ISSUE #133: Add index.ts with export for all bindings.
Hi

Would it be possible to also generate an `index.ts` inside the `bindings` directory?

## example

When I have the following structs:

```rust
#[derive(Serialize, Debug, TS)]
#[ts(
    export,
    export_to = "bindings/admin/login-response.ts",
    rename = "AdminLoginResponse"
)]
struct LoginResponse {
    access_token: String,
    expires_in: i64,
    token_type: String,
}

#[derive(Deserialize, Validate, Debug, TS)]
#[ts(
    export,
    export_to = "bindings/admin/login-request.ts",
    rename = "AdminLoginRequest"
)]
struct LoginRequest {
    #[validate(email(message = "invalid email"))]
    email: String,
    #[validate(length(min = 8))]
    password: String,
}
```

I would like to have a `index.ts` generated inside the `bindings` directory, that looks like:

```typescript
export * from './admin/login-request';
export * from './admin/login-response';
```

So I could add the following configuration to my `tsconfig.json`:

```json
{
  ...
  "compilerOptions": {
    ...
    "paths": {
      "@my-project/backend": ["backend/bindings/index.ts"],
    }
  }
}
```

this allows me to import all the exported types in my typescript code like this:

```typescript
import { AdminLoginRequest, AdminLoginResponse } from '@my-project/backend';
```

escritorio-gustavo:
Hey @NyxCode! This is a very rough first draft for the CLI tool we've been thinking about. It already handles all of our feature flags and the `TS_RS_EXPORT_DIR` environment variable

Closes #133 

## Todo
- [x] Find a way to only run `ts-rs` tests through this CLI
    -  Done by gating the generation of tests
        - Added a feature flag called `export` to `ts-rs` that re-enables test generation.
        - This feature is always enabled when using this CLI
- [x] Explore a way of generating an `index.ts` file to reexport all the generated types
    - During export, generate a metadata file to list all types and where they are exported to
        - In the CLI, use the metadata file to create the index.ts file
        - Gate the generation of this file behind a feature  flag named `generate-metadata`
- [x] Explore a way to merge all exports into a single file
    - Currently being done by reading each file listed in the metadata and copying its data over to `index.ts`, then deleting the original
- [ ] Discuss other features of the CLI
    - A configuration file so the user doesn't have to pass the flags every time?
        - How should the flags interact with the config file?
            - Should they be mutually exclusive?
            - Should they coexist with flags overriding the file?
            - Should there be a config flag to pass the path to the file?
        - An `init` subcommand to help the user create said config file?
    - Whatever else this CLI needs to do that isn't listed here
- [ ] When we consider it ready, it would be interesting to also make it available through `cargo binstall`

## Feature wishlist
- [ ] Customize representations (e.g `number` vs `bigint`) (@NyxCode) (being worked on in #334)
- [ ] Load configuration from configfile (@NyxCode) (being worked on in #334)
- [x] (maybe) Combine multiple types in one file? (@NyxCode) (awaiting review in #316)
- [x] (maybe) generate `index.ts` file, re-exporting types (@NyxCode) (See explanation above)

## Checklist

- [ ] I have followed the steps listed in the [Contributing guide](https://github.com/Aleph-Alpha/ts-rs/blob/main/CONTRIBUTING.md).


NyxCode:
Awesome!  

For `ts-rs`, we pretty often do *technically* breaking releases, but for most users, they are actually drop-in (unless e.g they implement `TS` manually). For the CLI, we probably want to get that right the first time, so I suggest we carefully plan what we want to support.

I've got a lot of open questions and ideas, but let me start with these two:

- Should the library be still usable without the CLI? I think it'd be a good idea to keep `cargo test` working, or maybe require `cargo test -- export_bindings`, but to keep it usable without CLI.
- Do we want to enable the `*-impl` features through the CLI? If we do that, then `#[derive(TS)]` must be a no-op normally, or code will fail to compile. Even then, if `#[derive(TS)]` only does something when the CLI is invoked, errors will only show up at that point. As far as I can tell, it'd make more sense for users to enable these features in the `Cargo.toml`.

NyxCode:
> - [ ] Explore a way of generating an `index.ts` file to reexport all the generated types
    - We could, after `cargo test` finishes, walk the output directory and create an index.ts containing `export * from 'file'`

That'd certainly work. Alternatively, the tests could just output the typescript (or/and metadata) to stdout, and the CLI could maybe combine them, or write the files (and an `index.ts`) itself.

escritorio-gustavo:
> * Do we want to enable the `*-impl` features through the CLI? If we do that, then `#[derive(TS)]` must be a no-op normally, or code will fail to compile. Even then, if `#[derive(TS)]` only does something when the CLI is invoked, errors will only show up at that point. As far as I can tell, it'd make more sense for users to enable these features in the `Cargo.toml`.

You're right, the `*-impl` features should not be in the cli

> * Should the library be still usable without the CLI? I think it'd be a good idea to keep `cargo test` working, or maybe require `cargo test -- export_bindings`, but to keep it usable without CLI.

I think it should, for that we could either merge #253 or add a cargo feature that blocks test generarion, so users would do `cargo test --features ts-rs/export` or something like that

escritorio-gustavo:
> Alternatively, the tests could just output the typescript (or/and metadata) to stdout, and the CLI could maybe combine them, or write the files (and an `index.ts`) itself.

We'd have to find a way to separate what we want from `cargo test`'s normal output (`test result: ok. 1 passed; 0 ...`)

escritorio-gustavo:
> I like that!

Awesome! I'll work on that later today, what do you think the feature should be called?

NyxCode:
> Awesome! I'll work on that later today, what do you think the feature should be called?

Oh, I'm terrible with names. But `export` seems reasonable :D

NyxCode:
> Do you think we should gate only the test generation behind the feature or should we make the whole #[derive(TS)] a no-op?
I considered both, but I think making `#[derive(TS)]` a no-op would be pretty intrusive, and potentially break lots of code (custom setups for exporting, custom TS impls with `: TS` trait bounds, etc.)

escritorio-gustavo:
Makes sense! I added an extra check with the `cfg!()` macro to verify the feature before generating the test

escritorio-gustavo:
> * [ ]  Customize representations (e.g `number` vs `bigint`)

Maybe we should revisit #94 to have a better idea of how to handle this

NyxCode:
Getting [#49359](https://github.com/rust-lang/rust/issues/49359) would be pretty cool - we could parse the JSON output, and do interesting stuff with it. 

```json
$ cargo +nightly t --features ts-rs/export -- -Z unstable-options --format json 2>/dev/null
{ "type": "suite", "event": "started", "test_count": 10 }
{ "type": "test", "event": "started", "name": "export_bindings_complexenum" }
{ "type": "test", "event": "started", "name": "export_bindings_complexstruct" }
{ "type": "test", "event": "started", "name": "export_bindings_gender" }
{ "type": "test", "event": "started", "name": "export_bindings_inlinecomplexenum" }
{ "type": "test", "event": "started", "name": "export_bindings_point" }
{ "type": "test", "event": "started", "name": "export_bindings_role" }
{ "type": "test", "event": "started", "name": "export_bindings_series" }
{ "type": "test", "event": "started", "name": "export_bindings_simpleenum" }
{ "type": "test", "event": "started", "name": "export_bindings_user" }
{ "type": "test", "event": "started", "name": "export_bindings_vehicle" }
{ "type": "test", "name": "export_bindings_complexstruct", "event": "ok" }
{ "type": "test", "name": "export_bindings_gender", "event": "ok" }
{ "type": "test", "name": "export_bindings_point", "event": "ok" }
{ "type": "test", "name": "export_bindings_role", "event": "ok" }
{ "type": "test", "name": "export_bindings_simpleenum", "event": "ok" }
{ "type": "test", "name": "export_bindings_vehicle", "event": "ok" }
{ "type": "test", "name": "export_bindings_series", "event": "ok" }
{ "type": "test", "name": "export_bindings_user", "event": "ok" }
{ "type": "test", "name": "export_bindings_complexenum", "event": "ok" }
{ "type": "test", "name": "export_bindings_inlinecomplexenum", "event": "ok" }
{ "type": "suite", "event": "ok", "passed": 10, "failed": 0, "ignored": 0, "measured": 0, "filtered_out": 0, "exec_time": 0.031168 }
{ "type": "suite", "event": "started", "test_count": 0 }
{ "type": "suite", "event": "ok", "passed": 0, "failed": 0, "ignored": 0, "measured": 0, "filtered_out": 0, "exec_time": 0.0001069 }
```

NyxCode:
[#49359](https://github.com/rust-lang/rust/issues/49359) seems [kinda close](https://internals.rust-lang.org/t/path-for-stabilizing-libtests-json-output/20163). [#50297](https://github.com/rust-lang/rust/issues/50297), on the other hand, seems far off. That's unfortunate, since we could probably do really cool stuff with it, without being too hacky.

NyxCode:
`--show-output` together with `--format json` seems especially cool. Any `println!` in the tests turns into 
```json
{ "type": "test", "name": "some_test", "event": "ok", "stdout": "hey\n" }
```

escritorio-gustavo:
> `--show-output` together with `--format json` seems especially cool. Any `println!` in the tests turns into
> 
> ```json
> { "type": "test", "name": "some_test", "event": "ok", "stdout": "hey\n" }
> ```

Now **that** would make it easier lol

NyxCode:
With the "generate an `index.ts`" feature in mind, the CLI needs *some* way to get a list of all TS types and their respective `.ts` files. Solutions I can think of are
- The tests communicate with the CLI by printing to stdout
  - That get's messy without `--format json`, I'm afraid
    - Maybe it'd be acceptable to require nightly for the CLI until that's stable, idk.
- The CLI parses the `.ts` files to figure out which types are in them
  - Lot of duplicated work there  
  - Alternative: We generate a comment in the first line of the `.ts` files, and just parse that.
- The tests don't generate `TS`, but `json` instead, and the CLI does the conversion to TS
- The tests generate an additional metadata file, which the CLI will parse

escritorio-gustavo:
> * The tests generate an additional metadata file, which the CLI will parse

I think this might be the best way to go. Use the mutex file lock to append to a metadata file all the types' names anc paths, then parse the metadata file to generate the index.ts file

NyxCode:
> I think this might be the best way to go. Use the mutex file lock to append to a metadata file all the types' names anc paths, then parse the metadata file to generate the index.ts file

I suspect that it might be simpler to write one metadata file per test, but i'd be happy to be convinced otherwise.

escritorio-gustavo:
> I suspect that it might be simpler to write one metadata file per test, but i'd be happy to be convinced otherwise.

I agree that this would probably be simpler, but we'd have to walk the output searching for these metadata files, so we could just walk the diretory reading the ts files instead

cli/src/main.rs
```diff
@@ -1,12 +1,15 @@
...
+
+mod path;
 
 #[derive(Parser, Debug)]
 struct Args {
     /// Defines where your TS bindings will be saved by setting TS_RS_EXPORT_DIR
-    #[arg(long, short)]
-    output_directory: Option<PathBuf>,
+    #[arg(long, short, default_value = "./bindings")]
+    output_directory: PathBuf,
```
escritorio-gustavo, line 12: I had to go back on obeying the environment variable if this argument isn't provided, otherwise there'd be no way to find the metadata file, as the CLI searches paths relative to the user's `pwd`.
Instead, we now always obey the cli (we could remove the default value to show the user that the CLI overrides the env variable)

cli/src/main.rs
```diff
@@ -33,26 +39,56 @@ macro_rules! feature {
...
+    color_eyre::install()?;
+
     let args = Args::parse();
 
     let mut cargo_invocation = std::process::Command::new("cargo");
 
+    let metadata_path = args.output_directory.join("ts_rs.meta");
+    if metadata_path.exists() {
+        std::fs::remove_file(&metadata_path)?;
+    }
```
escritorio-gustavo, line 52: Always delete previous runs' metadata file

cli/src/main.rs
```diff
@@ -33,26 +39,56 @@ macro_rules! feature {
...
+        let metadata = std::fs::read_to_string(&metadata_path)?
+            .lines()
+            .map(ToOwned::to_owned)
+            .collect::<std::collections::HashSet<_>>();
+
+        let index_path = args.output_directory.join("index.ts");
+
+        if index_path.exists() {
+            std::fs::remove_file(&index_path)?;
+        }
```
escritorio-gustavo, line 79: Always delete pre-existing index.ts as well

cli/src/main.rs
```diff
@@ -1,12 +1,15 @@
-use std::path::PathBuf;
+use std::{io::Write, path::PathBuf};
 
 use clap::Parser;
+use color_eyre::Result;
+
+mod path;
```
escritorio-gustavo, line 6: Had to duplicate `ts-rs/src/export/path.rs` to make absolute paths

ts-rs/src/export.rs
```diff
@@ -135,14 +135,27 @@ pub(crate) fn export_to<T: TS + ?Sized + 'static, P: AsRef<Path>>(
...
+            .to_string_lossy()
+            .bytes()
+            .chain([b'\n'])
+            .collect::<Box<[_]>>();
+        std::fs::OpenOptions::new()
+            .read(true)
+            .append(true)
+            .create(true)
+            .open(default_out_dir().join("ts_rs.meta"))?
+            .write_all(&bytes)?;
```
escritorio-gustavo, line 158: Currently the metadata file is just a list of the absolute paths of all exported files

escritorio-gustavo: If we plan to test this CLI against our test suite, there are a lot of naming collisions to resolve first

cli/src/main.rs
escritorio-gustavo: Should we make a `ts-rs-core` crate to share these types of functions?

escritorio-gustavo: Should we make a `ts-rs-core` crate to share these types of functions?

cli/src/main.rs
```diff
@@ -0,0 +1,99 @@
...
+            let mut index = std::fs::OpenOptions::new()
+                .create(true)
+                .append(true)
+                .open(index_path)?;
+
+            index.write_all(
+                b"// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.\n\n"
+            )?;
+            for file in metadata.iter() {
+                index.write_fmt(format_args!("export * from {file:?};\n"))?;
```
NyxCode, line 91: Ah, that's cool! Thought we had to name the types individually.  
As it turns out, since [TS 5.0](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#support-for-export-type), there's also `export type * from ..`, which we probably want to use here.

ts-rs/src/export.rs
```diff
@@ -145,17 +146,17 @@ pub(crate) fn export_to<T: TS + ?Sized + 'static, P: AsRef<Path>>(
...
 
-        let bytes = path::absolute(path.as_ref())?
-            .to_string_lossy()
-            .bytes()
-            .chain([b'\n'])
-            .collect::<Box<[_]>>();
+        let relative_path = T::output_path()
+            .ok_or_else(std::any::type_name::<T>)
+            .map_err(ExportError::CannotBeExported)?
+            .to_string_lossy();
```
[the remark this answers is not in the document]
NyxCode: That makes sense, though the metadata file isn't re-used anyway (even deleted).

ts-rs/src/export.rs
```diff
@@ -135,14 +135,27 @@ pub(crate) fn export_to<T: TS + ?Sized + 'static, P: AsRef<Path>>(
...
+
+        std::fs::OpenOptions::new()
+            .read(true)
+            .append(true)
+            .create(true)
+            .open(default_out_dir().join("ts_rs.meta"))?
+            .write_fmt(format_args!("./{relative_path}\n"))?;
```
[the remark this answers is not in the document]
escritorio-gustavo: > It's entirely possible that I'm overthinking this, but is this thread-safe as-is, or is some extra sync operation required here?

I was hoping that doing this inside the mutex lock would make it thread safe, but maybe we need to call `sync_data`? I don't actually know either

cli/src/main.rs
NyxCode: Hm, I guess we still need to name them individually, to avoid name collisions.  
In these cases, we'd want to emit something like `export type { A as A2, B } from "./a.ts";`

ts-rs/src/export.rs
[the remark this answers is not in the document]
NyxCode: > For most filesystems, the operating system guarantees that all writes are atomic: no writes get mangled because another process writes at the same time.

I think that means we'd be fine even without the lock as long as the write happens in just one syscall, which I think it does as long as we're not using a `BufWriter`.

cli/src/main.rs
escritorio-gustavo: We could also leave it to the user to resolve naming collisions by using different names for their types or using `#[ts(rename)]`. I think that if we rename the type it might be too confusing (i.e. user's type will be different than what they specified in Rust) and complex (i.e. we'd have to make sure that, if there are two types named `A`, we **always** rename the same one to `A1` and the other to `A2`, otherwise running the CLI twice in a row can change which type is which)

ts-rs/src/export.rs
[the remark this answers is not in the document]
escritorio-gustavo: I mean the path in `export * from "FILE"` in `index.ts`. An absolute path usually includes stuff specific to your machine like:
- Where you save your projects
- OS (a windows path is `C:\\foo\\bar.ts`, while a linux path is `/foo/bar.ts`)
- Username `/home/nyxcode/foo/bar.ts` or `C:\\Users\\escritoriogustavo\\foo\\bar`

cli/src/main.rs
NyxCode: Good point!

We'd want to disambiguate the names of both types, but I don't see a nice way of doing that. Even if we could, then adding a type later could cause a name conflict, which would break usages of the type which was already there.

So we could either detect the conflicts & show a nice error, or just generate invalid TS and let `tsc` tell the user what happened.

ts-rs/src/export.rs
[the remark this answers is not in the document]
NyxCode: Ah, thought you were referring only to the metadata file. Good point.

ts-rs/src/export.rs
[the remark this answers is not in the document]
escritorio-gustavo: Ah, so that's why we couldn't get it to break all the way back in #221! Nice

cli/src/main.rs
escritorio-gustavo: > So we could either detect the conflicts

That would be awesome if we managed to pull it off

NyxCode: I've got an idea, lemme see if that'd work out..

escritorio-gustavo: > That would be awesome if we managed to pull it 

The metadata file could have both the type's name as well as its path, separated by a `;` or `,` This would basically make it a csv file

If this is done, it's important that the type name comes first, as it is easier to use character like `,`/`;` to separate them as those are not valid for TS type names, and therefore can be safely assumed to not be part of the type name. That's harder to find with file paths

escritorio-gustavo: Then, in the CLI, we just check if any type name shows up more than once with a `HashMap<PathBuf, Vec<String>>` so we can say which types are repeated and to what files they were exported

NyxCode: Oh no, you were faster :D

escritorio-gustavo: I got it to work, might need some cleanup though

image

NyxCode: Nice! Might be cool to also have the `std::any::type_name` (or is there something better?) in the metadata file.  
Then, we could report the module path of the types as well.

escritorio-gustavo: `color_eyre` also reexports a trait that allows for coloring text:
image

escritorio-gustavo: > Nice! Might be cool to also have the `std::any::type_name` (or is there something better?) in the metadata file. Then, we could report the module path of the types as well.

That'd also be pretty helpful! I'll work on it right now.

escritorio-gustavo: image

NyxCode: That looks great, awesome!

NyxCode:
This all kinda makes me curious - How would it look like if we tried to do that without a CLI? Gotta play with that during the weekend!

cli/src/main.rs
```diff
@@ -64,7 +64,11 @@ fn main() -> Result<()> {
 
             index.write_all(NOTE)?;
 
-            for file in metadata.iter().map(|x| &x.1[0].export_path) {
```
escritorio-gustavo: This was only adding `export * from 'file'` for the first metadata entry, therefore, eliminating the naming collision from TypeScript's knowledge. But when there is a naming collision we want the index.ts file to be invalid

NyxCode:
> My guess is that it'd be pretty much impossible, at least the way we've implemented it so far, as it heavily relies on doing stuff after cargo test has finished

I suspect that's true. For something like `example/`, I think we could make it work with a `static` holding the current state, but as soon as there are multiple test executables, like in the `ts-rs` test suite, stuff get's very complicated.

cli/src/main.rs
escritorio-gustavo: Should we add a big error message at the end saying something like "Your index.ts file is not valid due to these naming collisions" or something or just let the user find out from `tsc`?

NyxCode: I think so! I dont see how a user could do anything useful with the invalid index file. Probably makes sense to not even write it to a file in that case.

escritorio-gustavo: image

cli/src/cargo.rs
```diff
@@ -31,11 +31,6 @@ pub fn invoke(args: &Args) -> Result<()> {
         } else {
             Stdio::piped()
         })
-        .stderr(if args.no_capture {
-            Stdio::inherit()
-        } else {
-            Stdio::piped()
-        })
```
escritorio-gustavo: This would capture this warning

image

cli/src/cargo.rs
```diff
@@ -46,6 +41,8 @@ pub fn invoke(args: &Args) -> Result<()> {
 
     if args.no_capture {
         cargo_invocation.arg("--").arg("--nocapture");
+    } else {
+        cargo_invocation.arg("--quiet");
```
escritorio-gustavo, line 45: This is to prevent cargo from logging progress messages ("Compiliing foo...", "Running unittests Bar" etc)

cli/src/args.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    /// Generates an index.ts file in your --output-directory that re-exports all
+    /// types generated by ts-rs
+    #[arg(long = "index")]
+    pub generate_index_ts: bool,
+
+    /// Generates only a single index.ts file in your --output-directory that
+    /// contains all exported types
+    #[arg(long = "merge")]
+    pub merge_files: bool,
```
gustavo-shigueo, line 34: Is this even necessary anymore? Since #316 you can just set all `#[ts(export_to = "...")]` to the same file and since #399, you can use a `const` to avoid any possible typos. These two features make this flag quite redundant

PR #306, opened 2024-04-11
Recusion Limit - Round 2

NyxCode:
## Goal

Determine when recusion limit errors still happen, reproduce and possibly fix them

## Changes
- Remove `TypeList`, move `TypeVisitor` to `lib.rs`
  - `dependency_types() -> TypeList` got changed to `visit_dependencies(v: &mut impl TypeVisitor)`
  - `generics() -> TypeList` got changed to `visit_generics(v: &mut impl TypeVisitor)`
  - codegen was adjusted accordingly

This is again a *technically* breaking change for anyone interacting with the `TS` trait directly.

## Checklist


NyxCode:
The "chain" of types `struct T000(T001)`, `struct T001(T002)`, ..., `struct T082(T083)`, `struct T083` manages to still run into the error.  

If there are only 83 instead of 84 types, then everything works, so that seems to be the limit currently.

```
error: reached the recursion limit while instantiating `<((), (PhantomData<T083>,)) as TypeList>::for_each::<...>`                                                                                                                                     
   --> C:\Users\morit\Coding\ts-rs\ts-rs\src\typelist.rs:102:9
    |
102 |         self.0.for_each(v);
    |         ^^^^^^^^^^^^^^^^^^
    |
note: `<(A, B) as TypeList>::for_each` defined here
   --> C:\Users\morit\Coding\ts-rs\ts-rs\src\typelist.rs:101:5
    |
101 |     fn for_each(self, v: &mut impl TypeVisitor) {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: the full type name has been written to 'C:\Users\morit\Coding\ts-rs\target\debug\deps\recursion_limit-57ec16690c4b1d02.long-type.txt'

error: could not compile `ts-rs` (test "recursion_limit") due to 1 previous error
```

NyxCode:
New codegen for 
```rs
#[derive(TS)]
#[ts(export, export_to = "generics/")]
struct Generic<T>
where
    T: TS,
{
    value: T,
    values: Vec<T>,
}
```
looks like this:
```rs
fn visit_generics(v: &mut impl  TypeVisitor) where Self: 'static, {
    v.visit::<T>();
    <T as TS>::visit_generics(v);
}
fn visit_dependencies(v: &mut impl  TypeVisitor) where Self: 'static, {
    <T as  TS>::visit_generics(v);
    <Vec<T> as  TS>::visit_generics(v);
    v.visit::<T>();
    v.visit::<Vec<T>>();
}
```

NyxCode:
Originally, I thought that having a `TypeList` would be nice. But it turned out that the only thing we ever do with it is to call `.for_each`.  
These changes basically put the `for_each` method directly into `TS`. `T::visit_dependencies(v)` is the same as `T::dependency_types().for_each(v)`, and `T::visit_generics(v)` is the same as `T::generics().for_each(v)`.

macros/src/deps.rs
```diff
@@ -79,26 +79,24 @@ impl Dependencies {
...
     fn to_tokens(&self, tokens: &mut TokenStream) {
-        let crate_rename = &self.crate_rename;
         let lines = self.dependencies.iter();
 
-        tokens.extend(quote![{
-            use #crate_rename::typelist::TypeList;
-            ()#(#lines)*
-        }]);
+        tokens.extend(quote![
+            #(#lines;)*
```
escritorio-gustavo, line 85: I love that there is no longer a need for the unit value before expanding `lines`

ts-rs/src/export.rs
```diff
@@ -85,7 +84,7 @@ mod recursive_export {
             out_dir,
             error: None,
         };
-        T::dependency_types().for_each(&mut visitor);
+        T::visit_dependencies(&mut visitor);
```
escritorio-gustavo, line 87: This definitely looks better than before. Nice!

ts-rs/src/lib.rs
```diff
@@ -423,15 +421,15 @@ pub trait TS {
     fn inline_flattened() -> String;
 
     /// Returns a [`TypeList`] of all types on which this type depends.
-    fn dependency_types() -> impl TypeList
+    fn visit_dependencies(_: &mut impl TypeVisitor)
```
escritorio-gustavo, line 424: I imagine this entirely kills the recursion problem caused by `TypeList` right?

escritorio-gustavo (approved):

NyxCode:
The `Generic` struct above nicely illustrates why it was easier to just get rid of `TypeList`.  
With the old `generics`, I wanted to generate
```rs
fn generics() -> impl TypeList where Self: 'static {
    struct List;
    impl TypeList for List {
        fn for_each(&self, v: &mut impl TypeVisitor) {
            v.visit::<T>();
            T::generics().for_each(v);
        }
    }
    List
}
```

Here, the `v.visit::<T>()` fails to compile, since the type parameter `T` can't be used within the `impl TypeList`. The workaround for this would be kinda complicated.

ts-rs/src/lib.rs
NyxCode: Jup, exactly! There's no huge tuple anymore for which the compiler has to figure out if it implements `TypeList`.

ts-rs/src/lib.rs
```diff
@@ -630,11 +635,13 @@ macro_rules! impl_tuples {
...
             }
-            fn dependency_types() -> impl TypeList
+            fn visit_dependencies(v: &mut impl TypeVisitor)
             where
                 Self: 'static
             {
-                ()$(.push::<$i>())*
+                $(
+                    v.visit::<$i>();
+                )*
```
NyxCode, line 644: Ha! Found a bug here!

ts-rs/src/lib.rs
```diff
@@ -630,11 +633,14 @@ macro_rules! impl_tuples {
...
-            fn dependency_types() -> impl TypeList
+            fn visit_generics(v: &mut impl TypeVisitor)
             where
                 Self: 'static
             {
-                ()$(.push::<$i>())*
+                $(
+                    v.visit::<$i>();
+                    <$i>::visit_generics(v);
+                )*
```
NyxCode, line 643: That should fix the tuple issue. There's no `visit_dependencies` right now - As far as I can tell, that'd only be relevant if `inline` did anything - for example inlining all contents of the tuple, like we do for `Vec`.

escritorio-gustavo:
Due to this PR removing trait functions that return `impl Trait`, if we make `parse_attrs` and `parse_serde_attrs` `pub(crate)` instead of just `pub`, it can lower the MSRV all the way down to 1.63.0!

escritorio-gustavo:
https://github.com/Aleph-Alpha/ts-rs/blob/d1223739632db2382eb419dc67e8cdac95d6430b/macros/src/utils.rs#L100
```suggestion
pub(crate) fn parse_attrs<'a, A>(attrs: &'a [Attribute]) -> Result<A> 
```

https://github.com/Aleph-Alpha/ts-rs/blob/d1223739632db2382eb419dc67e8cdac95d6430b/macros/src/utils.rs#L116
```suggestion
pub(crate) fn parse_serde_attrs<'a, A>(attrs: &'a [Attribute]) -> Serde<A> 
```

https://github.com/Aleph-Alpha/ts-rs/blob/d1223739632db2382eb419dc67e8cdac95d6430b/ts-rs/Cargo.toml#L18
```suggestion
rust-version = "1.63.0" 
```

macros/src/utils.rs
```diff
@@ -97,7 +97,7 @@ pub fn raw_name_to_ts_field(value: String) -> String {
 }
 
 /// Parse all `#[ts(..)]` attributes from the given slice.
-pub fn parse_attrs<'a, A>(attrs: &'a [Attribute]) -> Result<A>
+pub(crate) fn parse_attrs<'a, A>(attrs: &'a [Attribute]) -> Result<A>
```
escritorio-gustavo, line 100: In lower versions of Rust, this would complain about leaking a crate-private trait

NyxCode:
Merging this now. Not yet sure if we should publish a 9.0, or wait for the CLI. 
I guess publishing more often doesn't hurt, right?

escritorio-gustavo:
> Merging this now. Not yet sure if we should publish a 9.0, or wait for the CLI. I guess publishing more often doesn't hurt, right?

I don't mind publishing a 9.0, especially with the bug mentioned in #317 being fixed in this PR. The CLI can wait untill version 10

PR #312, opened 2024-04-19
i/usize types should generate bigint when in 64-bit systems

### LINKED ISSUE #311: `usize` and `isize` should be typed as `bigint` on 64-bit architectures
This would align with how, for example, `u32` and `u64` are currently typed.

escritorio-gustavo:
## Goal

Export `bigint` for `usize`, `isize` and their `NonZero` counterparts when in a system with 64-bit pointers
Closes #311 

## Checklist


NyxCode:
I'm really unsure about this.  
For the use-case I most often have in mind - JSON APIs - this would be super weird, since there's no correlation there.

Am I missing something here?

escritorio-gustavo:
This is kind of part of the debate going on in #94 about number/bigint. Since we consider numbers with 64+ bits as bigints #311 argues we should also consider i/usize as bigints on systems with 64 bit pointers

escritorio-gustavo:
This does affect JSON APIs, the creator of #311 commented that his API attempted to send a `usize` to the frontend, which failed to parse it because JS `number`s can't handle `usize`s, causing a runtime error

PR #316, opened 2024-05-10
feat: Allow multiple types to set `#[ts(export_to = "...")]` to the same file

### LINKED ISSUE #59: Combine exports into single file
I think that any exports within the same module should be combined into a single file, rather than splitting them up into separate files. I feel as though this should be the default, but perhaps a directive at the top of the module could indicate whether or not this happens?

escritorio-gustavo:
## Goal

Allow for multiple types to be exported to the same file by editing the target file when it is encountered by the second time
Closes #59

## Changes

Created a `HashMap` mapping file paths to a set of type identifiers and using it to decide whether or not the file should be created from scratch or edited

## Checklist

    - Thanks @martpie for helping test this feature


arslnb:
+ 1, looking forward to this!

escritorio-gustavo:
This feature is pretty much done, but I'd still like to make sure it doesn't break anything. If you have any projects that would benefit from this, would mind testing out this branch and checking if your resulting TypeScript is correct?

gustavo-shigueo:
Hey @NyxCode! Can you check this one out when you have some time?

martpie:
https://github.com/martpie/museeks/pull/776 is a test PR, all typings seem to be exported in a single file, in my TypeScript codebase (so outside my Rust codebase), which is exactly what I was looking for :)

Great job :)

gustavo-shigueo:
Thank you so much @martpie! It's awesome to see this working in an actual project with a lot of types

> I am one of those who commit their generated files to their codebase

This feature exports the types in a non-eterministic order, so you may see `index.ts` in your git diff everytime you run `ts-rs`, even if you don't actually change your types, I hope this isn't a problem

martpie:
My only nit: there seems to be more blank lines than needed ;)

gustavo-shigueo:
> My only nit: there seems to be more blank lines than needed ;)

I see, it was supposed to have only one blank line between types instead of two

gustavo-shigueo:
Hey @martpie, did this resolve the double blank line issue in your project?

gustavo-shigueo:
Yeah, I figured this could be a problem, I'll try to work on a way to sort the types alphabetically

gustavo-shigueo:
@martpie this new commit now sorts both imports and type declarations, so it should now be deterministic

martpie:
It's now stable :) 

but fyi:

- the sorting is done including the docstring I guess? Some types at the top because they have a comment starting with `/*`, but if the name of the type was used, they would not be sorted in this location.
- there are two `// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
` at the top

gustavo-shigueo:
> the sorting is done including the docstring I guess? Some types at the top because they have a comment starting with /*, but if the name of the type was used, they would not be sorted in this location.

Oof, didn't even think of that. Currently the sorting takes the whole declaration, including the docs

> there are two `// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.  at the top`

That's odd... I didn't manage to get this locally, but then again, my test just has 3 types, maybe I need more to trigger it

gustavo-shigueo:
I have just fixed the type sorting to consider only the type name (if the user has "export type " in the JSDoc, the sorting will unfortunately include that, but there's not much I can do about it)

I have also pushed an attempt at fixing the duplicated comment at the top of the file

martpie:
The sorting seems to be fixed, but I still see the header generated comment twice. Not a big deal, I'm just sharing what I see ;)

martpie:
There also seems to be some line breaks added here and there in the docstrings, but I'm really just nitpicking at this point (left is generated, right is the source)

Screenshot 2024-07-25 at 12 46 19

gustavo-shigueo:
> if the user has "export type " in the JSDoc, the sorting will unfortunately include that, but there's not much I can do about it

Slightly improved this by getting the last `"export type "`, now it'll only break if a field's JSDoc has that in it

gustavo-shigueo:
> There also seems to be some line breaks added here and there in the docstrings, but I'm really just nitpicking at this point (left is generated, right is the source)
> 
> Screenshot 2024-07-25 at 12 46 19

Did this happen with the version from crates.io? The docs feature was made with `///` comments in mind, so maybe it just has a bug dealiing with `/** */` comments.

gustavo-shigueo:
> The sorting seems to be fixed, but I still see the header generated comment twice. Not a big deal, I'm just sharing what I see ;)

I still have no idea how this is happening, but I really wanna get rid of this bug

gustavo-shigueo:
I copied over all of your exported types from martpie/museeks#776 into my local test file and that successfully reproduced the duplicated header error.

Now to getting it fixed... hopefully

gustavo-shigueo:
My guess is the problem is related to the way I look for `import` statements somehow breaking when there aren't any

gustavo-shigueo:
> My guess is the problem is related to the way I look for `import` statements somehow breaking when there aren't any.

Yup! `.trim_start_matches(NOTE)` fails in this case because the the trailing `\n` from the note is missing in the `<new | original>_header` variables

ts-rs/src/export.rs
```diff
@@ -158,9 +158,9 @@ fn export_and_merge(path: PathBuf, type_name: String, generated_type: String) ->
         .write(true)
         .open(&path)?;
 
-    let file_len = file.metadata()?.len() as usize;
+    let file_len = file.metadata()?.len();
 
-    let mut original_contents = String::with_capacity(file_len);
+    let mut original_contents = String::with_capacity(file_len as usize);
```
gustavo-shigueo, line 163: This change does nothing, it was leftover from when I was testing something else

martpie:
I can confirm the double header issue is fixed :]

gustavo-shigueo:
> Did this happen with the version from crates.io? The docs feature was made with `///` comments in mind, so maybe it just has a bug dealiing with `/** */` comments.

The way we generate the doc string is:
```rs
format!("/**\n{}\n */\n", lines.join("\n")),
```

This meas every JSDoc string is guaranteed to start with `/**\n` and end with `\n */\n`. Which explains the extra lines in your `PlayList` type

## Note
`lines` in the snippet represents a `Vec<String>`, where each item is the content of a single `#[doc = "..."]` attribute.

This works great with `///` comments, as the compiler converts each of these to its own `#[doc = "..."]` attribute, where `...` is replace with the contents of the comment.

However, a `/***/` comment is replaced with a single `#[doc = "..."]` attribute, where `...` is replaced with every single character between the second and third `*` characters, including any leading or trailing `\n`.

So
```rs
/**
 * Scan progress
 */
```
is replaced with `#[doc = "\n * Scan progress\n "]`. This leading `\n`, combined with our default `\n` + `"* "` before the `#[doc] contents`, causes the weird blank line

gustavo-shigueo:
I think we can fix it such that your `Scan progress` comment will render identically to how it is in Rust, but the track comment will probably remain as is

gustavo-shigueo:
@NyxCode I think this PR should be ready for merging, do you have anything you'd like to be changed here beforehand?

gustavo-shigueo:
> There also seems to be some line breaks added here and there in the docstrings, but I'm really just nitpicking at this point (left is generated, right is the source)

This should now be fixed too

Zerowalker:
Oh this would be neat.
Would be nice to have the option to do this globally as well.

For example I got a ton of types,
so having to do `#[ts(export_to = "...")]` on all is a bit messy.
So if possible overriding it to say "export all to X" or something would be neat.

Though I understand that might be beyond what this is attempting to do,
but just throwing it in there.

A great feature nonetheless :)

gustavo-shigueo:
There is `TS_RS_EXPORT_DIR`, which changes the directory the types are exported to. When used, all types will be exported to the given directory, and each type will be its own file, without needing to use `#[ts(export_to = "...")]`

martpie:
Any idea when this will be officially released? I know I can use a pointer to a github commit, but still ;)

gustavo-shigueo:
I think this feature will go into version 10 due to some breaking changes in other PRs, but I have to check with @NyxCode about that

PR #319, opened 2024-05-17
Testing with --no-default-features didnt seem to work in CI

NyxCode:


NyxCode:
@escritorio-gustavo Could you take a quick peek at this? 

As far as I can tell the CI ran `cargo t --no-default-features` only in the root directory, where the `--no-default-features` argument does nothing. So, possibly, `default-features = false` has been broken for a while. Oops!

Biggest change I've made here is to remove almost all of the `#[cfg(feature = "serde-compat")]` gates - I only kept them where we actually need them (e.g for printing the warning - we don't have the termcolor dependency without `serde-compat`).

In the other places, I used `if cfg!(feature = "serde-compat")` instead. That makes a lot of things, especially imports, easier.
As you can see in the first commit, those were most of the issues - `crate::attr::Serde` was imported, but it was gated behind the feature flag.

Does this seem alright to you?

escritorio-gustavo:
>  those were most of the issues - `crate::attr::Serde` was imported, but it was gated behind the feature flag.

Oops, sorry about that! Thankfully #290 hasn't been published yet as it was made after the release of version 8.1.0, so hopefully no one has been affected

> Biggest change I've made here is to remove almost all of the `#[cfg(feature = "serde-compat")]` gates [...], I used `if cfg!(feature = "serde-compat")` instead.
> [...]
> Does this seem alright to you?

Sure! I'm fine with both

PR #324, opened 2024-05-24
Convert Chrono Types to Date Objects in TypeScript

skrilax91:
## Goal

This PR aims to improve the type conversion of Chrono types in Rust to TypeScript by transforming them into Date objects instead of strings. This enhancement provides a more accurate and practical representation of date and time data in TypeScript.

## Changes

To address this issue, I modified the type conversion logic for Chrono types such as Date, Time, and Datetime. Instead of converting these types to strings, they are now converted to Date objects in TypeScript. This involved updating the relevant functions and ensuring that the resulting TypeScript code uses the Date type appropriately.

## Checklist


escritorio-gustavo:
Hi @skrilax91, thanks for the PR! Unfortunately, we cannot change the types for `chrono`'s date values.

The types we use are based on what happens when `serde` serializes the data. `NaiveDate`, `Date`, `DateTime` etc all get serialized to `string`, therefore we cannot make `Date` the default type as that would cause problems for users getting a JS `string` despite the TS type saying it should be a `Date`

If you want `Date`s in your JS you have to either:
- Convert the serde given string into a `Date` with `new Date(string)`
- Create custom serialization logic to produce `Date` objects and use `#[ts(type = "Date")]`

PR #333, opened 2024-06-21
Use `std::path::absolute`

### LINKED ISSUE #332: Use std::path::absolute
`std::path::absolute` [got stabilized in 1.79.0](https://blog.rust-lang.org/2024/06/13/Rust-1.79.0.html#stabilized-apis)

Nope!

Use the stdlib's `std::path::absolute`

**TODOs**
- [ ] figure out if we want to do that right away, or wait a bit until 1.79.0 is widely adopted
- [ ] make sure our path::absolute does 1:1 the same thing

gustavo-shigueo:
## Goal

Replace our hand implemented absolute path generator with the one provided by the Rust std lib
Closes #332

## Checklist


gustavo-shigueo:
It appears that the `std::path::absolute` function only cleans absolute pats on windows

NyxCode:
Ah, that sucks. I vaguely recall that there also was some reason we couldn't use `std::fs::canonicalize`, so idk. - maybe we should just keep it as-is? 😅

gustavo-shigueo:
`std::fs::canonicalize` requires the path to already exist, so it can't be used to create files or directories

PR #334, opened 2024-06-21
proof-of-concept of type overrides using the CLI

NyxCode:
This PR implements a POC of how we could implement type overrides using the CLI.


example/ts-rs.toml
```diff
@@ -0,0 +1,5 @@
+[overrides]
+u64 = "number"
+i64 = "number"
+u128 = "number"
+i128 = "number"
```
NyxCode, line 5: A user creates a `ts-rs.toml` config file in their project

cli/src/cargo.rs
```diff
@@ -26,20 +27,25 @@ pub fn invoke(args: &Args) -> Result<()> {
...
             Stdio::piped()
         })
-        .env("TS_RS_EXPORT_DIR", path::absolute(&args.output_directory)?);
+        .env("TS_RS_EXPORT_DIR", path::absolute(&cfg.output_directory)?);
 
-    feature!(cargo_invocation, args, {
+    for (rust, ts) in &cfg.overrides {
+        let env = format!("TS_RS_INTERNAL_OVERRIDE_{rust}");
+        cargo_invocation.env(env, ts);
+    }
```
NyxCode, line 40: the CLI reads the config file, and sets an environment variable for each override

NyxCode:
I am in no way sure that this is the right way to implement this. Besides the implementation (are environment variables really the right choice here?), these overrides would apply to the whole dependency chain. That might be the right behavior, but I'm not sure.

cli/src/config.rs
```diff
@@ -0,0 +1,104 @@
...
+impl Config {
+    pub fn load() -> Result<Self> {
+        let mut cfg = Self::load_from_file()?;
+        cfg.merge(Args::parse());
+        cfg.verify()?;
+        Ok(cfg)
+    }
+
+    fn load_from_file() -> Result<Self> {
+        // TODO: from where do we actually load the config?
```
gustavo-shigueo, line 66: I think there should be a CLI flag like `--config <PATH>` and `./ts-rs.toml` or maybe `./ts-rs/config.toml` should be a default location if the flag isn't provided

cli/src/main.rs
```diff
@@ -5,45 +5,45 @@ use std::{
...
 const BLANK_LINE: [u8; 2] = [b'\n', b'\n'];
 const NOTE: &[u8; 109] = b"// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.\n";
 
+struct Cleanup<'a>(&'a Config);
+impl<'a> Drop for Cleanup<'a> {
+    fn drop(&mut self) {
+        _ = fs::remove_file(self.0.output_directory.join(FILE_NAME));
+    }
+}
+
```
gustavo-shigueo, line 27: Is there a specific reason for `Cleanup` to be its own struct rather than having `impl<'a> Drop for Config<'a>` as it was with `Args`?

NyxCode: I think there was, though I can't completely recall.  
Maybe I moved it from `Args` into `Cleanup` to be able to destructure `Args`, though we may be able to move that back into `Config` if you prefer that.

Definetely like the idea of using a `Drop` impl like that, nice idea!

gustavo-shigueo: > I think there was, though I can't completely recall.

Just found it, turns out that ´Deserialize´ moves data out of `Config` which isn't allowed if `Config` implements `Drop`

gustavo-shigueo:
> I am in no way sure that this is the right way to implement this. [...] (are environment variables really the right choice here?)

I like this implementation, and I'm not sure there even is another option to get `cargo test` to read the overrides without environment variables

> Besides the implementation [...], these overrides would apply to the whole dependency chain. That might be the right behavior, but I'm not sure.

I think this is the right behavior. If, for example, the user wants `u64` to be treated as `number`, they probably want it like that all the way up  the chain

cli/src/main.rs
gustavo-shigueo: I figured out a way past it, I changed the code to go back to using `Args` and changed `Config` so that it'll only be used to read the config file

gustavo-shigueo: On a completely unrelated note, I never realized just how slow my home computer is until I ran `cargo test` and the `recursion_limit` tests took over 3 minutes to run xD

gustavo-shigueo:
Hey @NyxCode! Is there anything else you'd like to do before merging this into the cli branch?

gustavo-shigueo:
Hey @NyxCode, one question I have about this: why do we need one env variable for each type override? Couldn't a single env variable work, like `TS_RS_INTERNAL_OVERRIDE=String,string;i64,number` (ofc the separators I used here are arbitrary and can be replaced by anything else, as long as neither TS or Rust consider it a valid character inside identifiers)

NyxCode:
Yeah, I just did what seemed the most straight-forward here. Both reading them from a file (& caching it in a global?) and merging them into one environment variable are certainly possible.

Issue #335, opened 2024-06-27
Feature request: allow method of including `[propName: string]: any` to type


murl-digital:
i have a struct with a generic i'm exporting as a typescript type, however due to this struct being a part of a library where the generic will be provided by the user, there's not really a good way to define a type. since it's flattened, what i'd want to do is include a ``[propName: string]: any`` to the end like the typescript docs recommend, but i don't see a clean way to do that. here's the closest i've been able to get:

```rs

#[derive(TS)]
pub struct AnyMarker {
    #[ts(type = "any")]
    foo: ()
}

impl Document for AnyMarker {
    fn identifier() -> &'static str {
        todo!()
    }

    fn title() -> &'static str {
        todo!()
    }

    fn fields() -> Vec<EditorField> {
        todo!()
    }

    fn validators(model: DataModel) -> HashMap<String, ValidatorFunction> {
        todo!()
    }
}

#[derive(Serialize, Deserialize, TS)]
#[ts(export)]
#[ts(concrete(D = AnyMarker))]
pub struct Item<D: Document> {
    #[serde(rename = "__sc_id")]
    pub id: String,
    #[serde(rename = "__sc_created_at")]
    pub created_at: DateTime<Utc>,
    #[serde(rename = "__sc_modified_at")]
    pub modified_at: DateTime<Utc>,
    #[serde(rename = "__sc_published_at")]
    pub published_at: Option<DateTime<Utc>>,
    #[serde(flatten)]
    pub inner: D,
}
```

NyxCode:
Hey!
I'm not sure I fully understand what you're trying to do yet. What TS would you like to get?

NyxCode:
We do not support generating these "indexable" types, since there is no real rust equivalent.

If you **really** need to have a `[propName: string]: any` field, then you could add that field with some TypeScript.
```ts
type Item = ItemData & { [propName: string]: any };
```

You could also force ts-rs to generate that for you:
```rust
#[derive(ts_rs::TS)]
#[ts(export)]
struct Item {
  id: String,
  #[ts(flatten)]
  props: Props,
}

struct Props;
impl ts_rs::TS for Props {
   type WithoutGenerics = Self;
   fn decl() -> String { unreachable!() }
   fn decl_concrete() -> String { unreachable!() }
   fn name() -> String { unreachable!() }
   fn inline() -> String { unreachable!() }
   fn inline_flattened() -> String { "{ [propsName: string]: any }".to_owned() }
}
```

[...]

PR #336, opened 2024-06-27
Allow for flattening of generic parameters

NyxCode:
## Goal
Generic parameters currently can't be flattened:
```rust
#[derive(TS)]
#[ts(export)]
struct Item<D> {
    id: String,
    #[ts(flatten)]
    inner: D,
}
```
Serde happily accepts this, and we should too.

The three test cases work, and they are compatible with what serde does with them. There may be edge-cases I currently can't think of, though.  
However, this change definitely wont break any existing code.

## Changes
Tiny change: we just implement `fn inline_flattened()` for the dummy generic types we generate, instead of throwing an error. 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 ### Features
 ### Fixes
 
+# 9.1.0
+### Features
+- Allow using `#[ts(flatten)]` on fields using generic parameters ([#336](https://github.com/Aleph-Alpha/ts-rs/pull/336))
+
+
 # 9.0.0
 
 ### Breaking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,8 @@
 ### Features
 ### Fixes
 
-# 9.1.0
-### Features
+# 9.0.1
+### Fixes
 - Allow using `#[ts(flatten)]` on fields using generic parameters ([#336](https://github.com/Aleph-Alpha/ts-rs/pull/336))
 
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -70,7 +70,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "swc_macros_common",
- "syn 2.0.66",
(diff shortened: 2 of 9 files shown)


gustavo-shigueo (approved):
LGTM

NyxCode:
damn, that was fast! :D

gustavo-shigueo:
> damn, that was fast! :D

I literally logged into GitHub as the PR was being created lmao

NyxCode:
I *think* this is what @murl-digital wanted in #335, though it might be unrelated if I misunderstood.

gustavo-shigueo:
> I _think_ this is what @murl-digital wanted in #335, though it might be unrelated if I misunderstood.

Looking thorugh the issue, I think this is what they need as well

murl-digital:
> I _think_ this is what @murl-digital wanted in #335, though it might be unrelated if I misunderstood.

i clarified what i wanted in the issue just in case, but i think this would fit my usecase (i'll have to check, though)

one potential issue is handling trait bounds, but i think i can just use a concrete type or something.

NyxCode:
> one potential issue is handling trait bounds, but i think i can just use a concrete type or something.

Since your definition of `struct Item` doesn't make use of the fact that `D: Document`, I'd consider if you can just remove that bound. Normally, you only need such a bound on a struct definition if you're using an associated type on `Document`.  
Even `std::hashmap::HashMap<K, V>` doesn't constrain `K` or `V`, but only the `impl<K: Hash> HashMap { ... }` block does.

NyxCode:
Hm, maybe this should be a patch release instead..

gustavo-shigueo:
> Hm, maybe this should be a patch release instead..

This would fit very well in a `9.1.0`, maybe we could even squeeze #316 in with it

gustavo-shigueo:
Hey @NyxCode, since this was release on crates.io, I'll make a corresponding GitHub release

Issue #186, opened 2023-12-28
Include `ObjectId` from `mongodb`


bennyhodl:
When I generate bindings with the `bson` feature it does not include the import in the generated files.

How can I get it to auto import?

escritorio-gustavo:
Hey @bennyhodl, there have been a lot of changes to the import logic recently, can you try in the current `main` and, if, the issue persists, send us a minimal reproduction?

escritorio-gustavo:
Closing this until a minimal repro is submitted

PR #340, opened 2024-07-15
feat: add bson ObjectId impl

ShaunSHamilton:
## Goal

The docs suggest the `bson-uuid-impl` feature flag implements types from `bson`. However, this is misleading, because it just implements `TS` for `bson::Uuid`.

On top of this, the library suggests there is a `bson` feature flag. However, enabling this does nothing. So, this pr adds an impl for `bson::oid::ObjectId` which is a particularly common type.

This PR does **not** expect to implement every exported type from `bson`. Hopefully, this will be done in future releases of `ts-rs`.

Related to https://github.com/Aleph-Alpha/ts-rs/issues/186

## Changes

- Added `bson::oid::ObjectId` impl under `bson` feature config.
- Changed lib docs about `bson` feature flag

## Checklist

## Notes

The docs suggest running `cargo +nightly fmt`. However, after running this, unrelated files were changed. So, those have **not** been pushed.

`bson` serializes `ObjectId` into `{ "$oid": <string> }`. However, this is likely **not** what users want in their types.


README.md
```diff
@@ -79,8 +79,9 @@ When running `cargo test`, the TypeScript bindings will be exported to the file
...
 | url-impl           | Implement `TS` for types from *url*                                                                                                                                                                       |
-| uuid-impl          | Implement `TS` for types from *uuid*                                                                                                                                                                      |
-| bson-uuid-impl     | Implement `TS` for types from *bson*                                                                                                                                                                      |
+| uuid-impl          | Implement `TS` for types from *uuid*
+| bson               | Implement `TS` for *bson::oid::ObjectId*                                                                                                                                                                  |
+| bson-uuid-impl     | Implement `TS` for *bson::Uuid*                                                                                                                                                                      |
```
gustavo-shigueo, line 84: Hey @ShaunSHamilton, thanks for the PR. This all looks good to me, but maybe it'd be a good idea to collapse both `bson` related features into a single cargo feature that that implements both `bson::Uuid` and `bson::oid::ObjectId`?

Unless of course this isn't possible (maybe it requires some feature to be enabled in `bson`? I never used this crate), if that's the case, just inform me and we can get this merged

ShaunSHamilton: 100%. The only reason I can see to not combine the two impls is if a consumer only needed one.

I am not sure if this would be a major change, though. Currently, both `bson` and `bson-uuid-impl` flags exist, however, only the `bson-uuid-impl` flag was documented, and the `bson` flag did nothing.

Shall I add the `bson::Uuid` impl to the `bson` flag, and leave the `bson-uuid-impl` untouched?

gustavo-shigueo: I see, so, the `bson` flag enables `bson::oid::ObjectId` and this doesn't require any features from `bson`, only the crate itself correct?

And `bson-uuid-impl` enables `bson::Uuid` which requires some `bson` with some feature enabled?

gustavo-shigueo: Also, it seems `bson-uuid-impl` was not added to `Cargo.toml`

gustavo-shigueo: Wait, as far as I could tell, both `bson::oid::ObjectId` and `bson::Uuid` are available in `bson` by default, I think it should be possible to just have a `bson-impl` feature to handle both

ShaunSHamilton: Whilst this works, and I am happy it is in, it would make more sense to prioritise a plain `bson` flag.

Are you happy with me opening a PR to?:
- Revert `ObjectId` inclusion in `bson-uuid-impl`
- Add `ObjectId` as well as all other default pub structs to `bson` flag

Ideally, that would go in before the next release, and `bson-uuid-impl` could be left as is.

gustavo-shigueo: If you do decide to create this PR, I think the flag should be called `bson-impl` as all other external crate support flags follow this `<crate>-impl` naming scheme. Also, since the `bson` crate doesn't hide `bson::oid::ObjectId` or `bson::Uuid` behind any feature flags, they can both just be moved into `bson-impl` and `bson-uuid-impl` can just be removed.

Also, that'll be a breaking change, so it'll only make it into the library in version 10, so it will require @NyxCode's approval, especially since I have no idea how `bson` works 😅

PR #346, opened 2024-08-07
Always use forward slash on import paths

### LINKED ISSUE #345: Dependency paths in Windows use back slash, though they should use the canonical forward slash
On windows, generated definitions are created with backslashes as opposed to forward slashes:

```ts
import type { MyType } from "..\\my\\windows\\path";

// should be:
import type { MyType } from "../my/windows/path";
```

Rust and typescript are smart enough to handle unix paths, and the difference creates some noise when working across windows and os x environments. Note: I believe this only happens when the dependency includes "..". It seems like there's a migration to a new path api in the works, so that might address this.

gustavo-shigueo:
## Goal

This PR changes import paths such that they will always use forward slash `/` as the path separator, even on Windows. It also adds a windows test run to the CI
Closes #345

## Changes

If `target_os = "windows"`, replace `\` with `/`

## Checklist

### DIFF
--- a/ts-rs/src/export.rs
+++ b/ts-rs/src/export.rs
@@ -323,7 +323,7 @@ fn generate_imports<T: TS + ?Sized + 'static>(
 
         writeln!(
             out,
-            "import type {{ {} }} from {:?};",
+            r#"import type {{ {} }} from "{}";"#,
             &dep.ts_name, rel_path
         )?;
     }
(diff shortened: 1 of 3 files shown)


ts-rs/src/export.rs
```diff
@@ -334,11 +334,19 @@ fn generate_imports<T: TS + ?Sized + 'static>(
...
-        Some(Component::Normal(_)) => format!("./{}", rel_path.to_string_lossy()),
+    let str_path = match rel_path.components().next() {
+        Some(Component::Normal(_)) => {
+            format!("./{}", rel_path.to_string_lossy())
+        }
         _ => rel_path.to_string_lossy().into(),
     };
 
+    let path = if cfg!(target_os = "windows") {
+        str_path.replace('\\', "/")
```
NyxCode, line 345: I dont think this is sufficient, since paths can contain \.

We might need to iterate over the path components, not sure though.

NyxCode (changes requested):

ts-rs/src/export.rs
gustavo-shigueo: > I dont think this is sufficient, since paths can contain \.

Wdym? Is a component allowed to contain `\`? This replaces every `\` for `/`

gustavo-shigueo: It seems `\` is allowed in Unix file/directory names, but this shouldn't be a problem, as the replace only happens if the user's OS is Windows, where `\` is not allowed.

I also suspect TS itself would not be very fond of a file containaing `\` in its name.

It could even be a reasonable idea to throw a compile time error if `#[ts(export_to)]` or `TS_RS_EXPORT_DIR` contain backslashes as that would probably produce unintended behavior and destroy portability, by having a path that can't exist on Windows. Of course, doing so would be another of those "*technically* a breaking breaking change, but probably won't affect anyone" versions

NyxCode: I think you're right, sorry about that!

I thought I remembered some weird rules for paths on windows, but i was mistaken.

NyxCode (approved):
thanks man! especially happy about the windows CI!

PR #351, opened 2024-09-07
Support for rust_decimal::Decimal

corvusrabus:
## Goal

Adds support for the currently most popular decimals crate` rust_decimal` ([23m downloads on crates.io](https://crates.io/crates/rust_decimal)). 

I think the crates `rust_decimal` and `bigdecimal` don't need `serde` themselves enabled as the users of this library will enable it by themselves if they need it.

## Checklist


corvusrabus:
@gustavo-shigueo sorry you were so fast with merging my last one that I had to open a new one for this :)

corvusrabus:
@gustavo-shigueo any chance this could be release `9.1` ?

gustavo-shigueo:
Don't worry about smol_str, I've updated the feature name on `main`

NyxCode:
How does this interact with the serde features of `rust_decimal`?

There's
- [serde-float](https://docs.rs/rust_decimal/latest/rust_decimal/#serde-float)
- [serde-str](https://docs.rs/rust_decimal/latest/rust_decimal/#serde-str)
- [serde-arbitrary-precision](https://docs.rs/rust_decimal/latest/rust_decimal/#serde-arbitrary-precision)
- [serde-with-float](https://docs.rs/rust_decimal/latest/rust_decimal/#serde-with-float)
- [serde-with-str](https://docs.rs/rust_decimal/latest/rust_decimal/#serde-with-str)
- [serde-with-arbitrary-precision](https://docs.rs/rust_decimal/latest/rust_decimal/#serde-with-arbitrary-precision)


We had some discussion about this a while back in #273. As far as I can tell, we cannot support this library here properly, and support within `rust_decimal` itself is needed.

corvusrabus:
> How does this interact with the serde features of rust_decimal?

You're completely right about this suspicion. I didn't remember that `rust_decimal` has the weird architecture of being allowed to change the serde serialization via feature flag. 

I don't see how to get arund this right now. We might in a hacky way serialize a Decimal via `serde_json::to_value` and see which type we get but I am not sure how to do that at compile time to deduce the types needed to implement `TS`.

Closing this for now and might reopen if I come up with a new idea.

Any chance you could release version 9.1 that includes the `smolstr` feature @NyxCode ?

martinbliss:
Would it be possible to revisit this? I'm using sea-orm's CLI to generate rust types which utilize `rust-decimal` so I have limited flexibility on changing to a different type, and I just need `strings` on the TS side. Would it be possible to support the vanilla use case of `rust-decimal::Decimal -> string` and add a disclaimer in the README for this approach not being appropriate for extra cases? @corvusrabus  @NyxCode 

I was about to open a PR just ilke this one when I decided to do a gut check and see if anyone tried to support this because I didn't see `rust-decimal` mentioned anywhere but I understand why this hasn't been added yet.

PR #357, opened 2024-09-19
Release v10.0.0

NyxCode:


NyxCode:
I'm still a bit nervous about #316. But I tried a few things to break it, and failed every time. So I guess we're good? 😄

gustavo-shigueo:
I think we're good, I couldn't break it either and @martpie tested it on his project [museeks](/martpie/museeks), where it seems to be working fine

gustavo-shigueo (approved):

martpie:
It's been working perfectly for a few weeks, and the output is stable, so that's perfect in my case :]

PR #366, opened 2024-11-09
Add `#[ts(optional)]` to struct

### LINKED ISSUE #364: Feature request: Easier way to output typescript optionals, IE `?: Thing` for rust `Option` types, rather than `Thing | null`

We're upgradiing our `ts-rs` from `7.1.1` -> `10.0.0`, but unfortunately this upgrade broke all of our `Option` types.

Beforehand, we could do this on a struct:

```rust
#[skip_serializing_none]
#[cfg_attr(feature = "full", ts(export))]
pub struct AdminPurgePersonView {
  admin: Option<Person>
}
```

And this would generate:

```typescript
export interface AdminPurgePersonView {
  admin?: Person
```

With the upgrade, the above now generates: 

`admin: Person | null`
  
This must be because ts-rs is now ignoring serde's `skip_serializing_none` directive.

Have ts-rs follow the above directive.

**Describe alternatives you've considered**

- Add `ts(optional)` to every single field, but this would be tedious and verbose.
- Change the default behavior of rust's options to match actual generated [typescript optionals](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#optional-properties) (IE `?`) , rather than outputting `thing | null`, which isn't the preferred way typescript deals with optionals.

You don't need to get to into it if you don't want, but what's the short version for why ts-rs outputs `thing | null` rather than proper typescript [optionals](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#optional-properties)?

gustavo-shigueo:
## Goal

Allow the use of `#[ts(optional)]` on struct definitions
Closes #364

## Checklist

### DIFF
--- a/macros/src/types/named.rs
+++ b/macros/src/types/named.rs
@@ -105,30 +105,18 @@ fn format_field(
     let ty = field_attr.type_as(&field.ty);
 
     let opt = match (struct_optional, field_attr.optional) {
-        (
-            opt @ Optional { optional: true, .. },
-            Optional {
-                optional: false, ..
-            },
-        ) => opt,
-        (_, opt @ Optional { optional: true, .. }) => opt,
-        (
-            opt @ Optional {
-                optional: false, ..
-            },
-            Optional {
-                optional: false, ..
-            },
-        ) => opt,
+        (opt @ Optional::Optional { .. }, Optional::NotOptional) => opt,
+        (_, opt @ Optional::Optional { .. }) => opt,
+        (opt @ Optional::NotOptional, Optional::NotOptional) => opt,
     };
 
-    let optional_annotation = if opt.optional {
+    let optional_annotation = if let Optional::Optional { .. } = opt {
         quote! { if <#ty as #crate_rename::TS>::IS_OPTION { "?" } else { "" } }
     } else {
         quote! { "" }
     };
 
-    let optional_annotation = if field_attr.optional.optional {
+    let optional_annotation = if let Optional::Optional { .. } = field_attr.optional {
         quote! {
             if <#ty as #crate_rename::TS>::IS_OPTION {
                 #optional_annotation
@@ -153,7 +141,7 @@ fn format_field(
             if field_attr.inline {
                 dependencies.append_from(&ty);
 
-                if opt.optional && !opt.nullable {
+                if let Optional::Optional { nullable: false } = opt {
                     quote! {
                         if <#ty as #crate_rename::TS>::IS_OPTION {
                             <#ty as #crate_rename::TS>::option_inner_inline().unwrap()
@@ -166,7 +154,7 @@ fn format_field(
                 }
             } else {
                 dependencies.push(&ty);
-                if opt.optional && !opt.nullable {
+                if let Optional::Optional { nullable: false } = opt {
                     quote! {
                         if <#ty as #crate_rename::TS>::IS_OPTION {
                             <#ty as #crate_rename::TS>::option_inner_name().unwrap()
--- a/macros/src/types/named.rs
+++ b/macros/src/types/named.rs
@@ -1,7 +1,7 @@
 use proc_macro2::TokenStream;
-use quote::quote;
-use syn::{Field, FieldsNamed, Path, Result};
-
+use quote::{quote, quote_spanned};
+use syn::{parse_quote, Field, FieldsNamed, Path, Result};
+use syn::spanned::Spanned;
 use crate::{
     attr::{Attr, ContainerAttr, FieldAttr, Inflection, Optional, StructAttr},
     deps::Dependencies,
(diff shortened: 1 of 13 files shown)


NyxCode:
Love the idea. Would also like to hear what @Nutomic and @dessalines think.  

I'm not yet sure if the re-using the same `#[ts(optional)]` on a struct is a good idea syntax-wise, but maybe it'll grow on me.

dessalines:
Excellent! Yep this would really gonna clean up our structs. Thx!

NyxCode:
@gustavo-shigueo A problem I see with this implementation is that *all* fields have to be `Option<T>`, otherwise this error is emitted: `error: `optional` can only be used on an Option<T> type`.

We could fix that by trying to figure out if a field is an `Option`, and if it isn't, using the default behaviour. But like the current handling of `#[ts(optional)]`, that's fragile, and breaks with type aliases.

 Maybe we can move that logic from the macro to the runtime, where we actually know if a type is an Option or not.

gustavo-shigueo:
> @gustavo-shigueo A problem I see with this implementation is that _all_ fields have to be `Option<T>`, otherwise this error is emitted: `error: 'optional' can only be used on an Option<T> type`.

Damn, I forgot about that

> We could fix that by trying to figure out if a field is an Option, and if it isn't, using the default behaviour. But like the current handling of `#[ts(optional)]`, that's fragile, and breaks with type aliases.

Yeah, that would probably not be a great idea

gustavo-shigueo:
I implemented a band-aid fix for the problem, but we should probably look into a better way to detect the `Option` type

gustavo-shigueo:
> Maybe we can move that logic from the macro to the runtime, where we actually know if a type is an Option or not.

We could add a `const IS_OPTION: bool = false` to the `TS` trait and set it to `true` in our `impl<T: TS> TS for Option<T>`. That shouldn't even be a breaking change since the trait provides a default value

gustavo-shigueo:
@NyxCode  I have implemented this idea for you to take a look, I also change `Optional` into an enum, this way it is impossible to represent the invalid `Optional { optional: false, nullable: true }`

gustavo-shigueo:
The downside of this implementation is that there is no compile time error for 
```ts
struct Foo {
    #[ts(optional)]
    foo: i32
}
```

Instead, this will panic when the user attempts to export their types. I don't quite like this because ts-rs usually works in a "If it compiles, you're good" way, but here we get something that compiles and doesn't work.

NyxCode:
Nice work! I have an idea how to get this compile-time checked, lemme see if that actually works though..

macros/src/types/named.rs
```diff
@@ -102,27 +102,35 @@ fn format_field(
...
+                #optional_annotation
+            } else {
+                panic!("`#[ts(optional)]` can only be used with the Option type")
             }
         }
-        Optional {
-            optional: false, ..
-        } => (&parsed_ty, ""),
+    } else {
+        optional_annotation
```
gustavo-shigueo, line 128: Another option is just remove this whole thing, this way there'll be no panic (still, no compiler error). What will happen is that `#[ts(optional)]` will just do nothing when not using an Option type

gustavo-shigueo:
> I'm not yet sure if re-using the same `#[ts(optional)]` on a struct is a good idea syntax-wise, but maybe it'll grow on me.

Good point, maybe at the struct level we should call it something like `optional_fields`

gustavo-shigueo:
> maybe at the struct level we should call it something like `optional_fields`

@NyxCode  I've applied this change, but if you think of a better name, feel free to push a commit here

gustavo-shigueo:
I also added the breaking change tag because this will break manual implementations of TS. As always, the feature we need to avoid that is unstable lol (rust-lang/rust#29661)

NyxCode:
I think I'm happy with the new name - nice!  
Besides docs, is there anything we need to get this merged?

gustavo-shigueo:
> Besides docs, is there anything we need to get this merged?

No, I think it's just the docs

NyxCode:
Should we infer `#[ts(optional_fields)]` when `#[serde(skip_serializing_if = "Option::is_none")]` and `#[serde(default)]` is present?  
I believe that would be correct, though possibly somewhat unpredictable, so I'm not sure if that's a good idea.

NyxCode:
What's the expected behavior with `#[ts(optional_fields)]` on a struct and `#[ts(type)]` on a field?
```rust
#[ts(optional_fields)]
struct X {
  a: Option<i32>, // a?: number
  #[ts(type = "string")]
  b: Option<SomethingElse>,
}
```
Right now, the `#[ts(type = "..")]` works as expected - but we still get the `?` for `b`: `b?: string`.

gustavo-shigueo:
> What's the expected behavior with `#[ts(optional_fields)]` on a struct and `#[ts(type)]` on a field?
> 
> ```rust
> #[ts(optional_fields)]
> struct X {
>   a: Option<i32>, // a?: number
>   #[ts(type = "string")]
>   b: Option<SomethingElse>,
> }
> ```
> 
> Right now, the `#[ts(type = "..")]` works as expected - but we still get the `?` for `b`: `b?: string`. I do suspect that `#[ts(type)]` should override that, resulting in just `b: string`.

Nice catch!, I've changed `format_field` to handle `#[ts(type = "...")]` in isolation from other attibutes (except for `rename` and `rename_all`), I also made it so you cannot use `#[ts(type = "...")]` and `#[ts(optional)]` on the same field, as `#[ts(optional)]` would have no effect

NyxCode (approved):
Hey @gustavo-shigueo! I am - and have been - very busy lately. Sorry for being less active here.  

Nonetheless, I'd like to get this merged, and release a new version in a new PR.

The only thing missing here is a note in the changelog, I think. Is there something else you'd like to change/note/add?

gustavo-shigueo:
> Hey @gustavo-shigueo! I am - and have been - very busy lately. Sorry for being less active here.   
> Thank you for keeping an eye on everything!

Hey @NyxCode, I hope you're doing well! No problem, I've also been pretty busy lately, I've been keeping an eye on the issues but haven't written any code in a while.

> Nonetheless, I'd like to get this merged, and release a new version in a new PR.
> 
> The only thing missing here is a note in the changelog, I think.

Awesome! I'll make sure to add it tomorrow morning, thanks!

gustavo-shigueo:
Updated the changelog, I also ran `cargo msrv verify` to check if our `rust-version` was correct and it turns out the MSRV is now `1.78.0`

gustavo-shigueo:
> Updated the changelog, I also ran `cargo msrv verify` to check if our `rust-version` was correct and it turns out the MSRV is now `1.78.0`

I'll see if I can make a GH workflow this weekend to check that for us in the future

dessalines:
Would it be possible to do a minor release sometime soon (that has this)?

gustavo-shigueo:
This is a breaking change for anyone manually implementing `TS`, so this must be in a major release

gustavo-shigueo:
I think it could be possible to have a major release about now, the only PR that I think is worth looking into before hand is #376, which is waiting for @NyxCode's review, nad maaaaybe we could look into having a pre-release version of the CLI, but that's unlikely

NyxCode:
@gustavo-shigueo I've raised the permissions issue with the org. They reshuffled the management of it, but I hope they'll finally get around to it after the easter holidays. Again, sorry about that.

I'll merge the deno PR & cut a release early next week.

gustavo-shigueo:
> @gustavo-shigueo I've raised the permissions issue with the org. They reshuffled the management of it, but I hope they'll finally get around to it after the easter holidays. Again, sorry about that.

Thanks for looking into it! No need to apologize, though. This isn't on you at all

> I'll merge the deno PR & cut a release early next week.

Awesome!

gustavo-shigueo:
> I hope they'll finally get around to it after the easter holidays

Hey @NyxCode, just checking if have you've heard anything from them

NyxCode:
Hey @gustavo-shigueo!
I've just checked again, and I think they've fixed it! You should have gotten an invitation, probably per email?

gustavo-shigueo:
Hey @NyxCode, yeah, I just received it, thanks!

NyxCode:
Yay! Man, that was a journey. Sorry that it took so long.

gustavo-shigueo:
Don't worry about it

dessalines:
Happy to say we'll be using this feature soon in lemmy: https://github.com/LemmyNet/lemmy/pull/5777

Thx again!

PR #367, opened 2024-11-11
compile-time checked `#[ts(optional)]`

NyxCode:
Builds on top of #366.

- Get's rid of all string/path manipulation in the macro
  - As a consequence, it's impossible to break this with type aliases in all cases. 
- Produces a compile-time error when `#[ts(optional)]` is used on a non-`Option` field. The error is not great yet - it's probably possible to improve this though.
```rust
    #[ts(optional = nullable)]
    d: i32,
```
```
error[E0308]: mismatched types
  --> ts-rs/tests/integration/optional_field.rs:94:10
   |
94 | #[derive(TS)]
   |          ^^
   |          |
   |          expected `PhantomData<Option<_>>`, found `PhantomData<i32>`
   |          expected due to this
   |
   = note: expected struct `PhantomData<std::option::Option<_>>`
              found struct `PhantomData<i32>`

```

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -80,6 +80,7 @@ impl DerivedTS {
         quote! {
             #impl_start {
                 #assoc_type
+                type OptionInnerType = Self;
 
                 fn ident() -> String {
                     #ident.to_owned()
@@ -156,6 +157,7 @@ impl DerivedTS {
                 }
                 impl #crate_rename::TS for #generics {
                     type WithoutGenerics = #generics;
+                    type OptionInnerType = Self;
                     fn name() -> String { stringify!(#generics).to_owned() }
                     fn inline() -> String { panic!("{} cannot be inlined", #name) }
                     fn inline_flattened() -> String { stringify!(#generics).to_owned() }
--- a/macros/src/types/enum.rs
+++ b/macros/src/types/enum.rs
@@ -24,7 +24,7 @@ pub(crate) fn r#enum_def(s: &ItemEnum) -> syn::Result<DerivedTS> {
     if let Some(attr_type_override) = &enum_attr.type_override {
         return type_override::type_override_enum(&enum_attr, &name, attr_type_override);
     }
- 
+
     if let Some(attr_type_as) = &enum_attr.type_as {
         return type_as::type_as_enum(&enum_attr, &name, attr_type_as);
     }
--- a/macros/src/types/named.rs
+++ b/macros/src/types/named.rs
@@ -111,7 +111,7 @@ fn format_field(
             quote! {{
                 use std::marker::PhantomData;
                 let actual: PhantomData<#ty> = PhantomData;
-                let must: PhantomData<Option<_>> = actual;
+                let must: PhantomData<std::option::Option<_>> = actual;
                 "?"
             }},
             nullable,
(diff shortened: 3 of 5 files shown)


NyxCode:
This is achieved (technically semver-breaking) by adding an associated type, `OptionInnerType` to the `TS` trait. This type is set to `Self` for all types, except for `Option<T>`, where it is set to `T´.

With that, the behaviour of `#[ts(optional)]` never errors by itself. The mechanism to produce an error if it's used on a non-`Option` field is completely separate.

This is done by emitting this code:
```rust
use std::marker::PhantomData;
let actual: PhantomData<#ty> = PhantomData;
let must: PhantomData<Option<_>> = actual;
```
If `#ty` is not an `Option`, that code fails to compile, and produces the error as seen above.

NyxCode:
Alternative way to provoke a compile-time error:
```rust
fn a<T>(_: std::option::Option<T>) {}
fn b<T>(x: #ty) { a(x) }
```

NyxCode:
Notably, what unfortunately does **not** work is something like this:
```rust
let true = <#ty as TS>::IS_OPTION;
```

NyxCode:
I massively improved the error. For 
```rust
#[derive(TS)]
#[ts(export, export_to = "optional_field/", optional)]
struct OptionalStruct {
    // `#[ts(optional)]` on a type that isn't `Option<T>` does nothing
    #[ts(optional = nullable)]
    d: i32,
}
```
we now get
```
error[E0277]: `#[ts(optional)]` can only be used on fields of type `Option`
   --> ts-rs/tests/integration/optional_field.rs:104:5
    |
104 |     #[ts(optional = nullable)]
    |     ^
    |
    = help: the trait `IsOption` is not implemented for `i32`
    = note: `#[ts(optional)]` was used on a field of type i32, which is not permitted
    = help: the trait `IsOption` is implemented for `std::option::Option<T>`
```

There is a weird note attached to the error though, which I don't know how to get rid of:
```
note: required by a bound in `<OptionalStruct as TS>::inline::check_that_field_is_option`
   --> ts-rs/tests/integration/optional_field.rs:94:10
    |
94  | #[derive(TS)]
    |          ^^ required by this bound in `check_that_field_is_option`
...
104 |     #[ts(optional = nullable)]
    |     - required by a bound in this function
    = note: this error originates in the derive macro `TS` (in Nightly builds, run with -Z macro-backtrace for more info)
```

gustavo-shigueo (approved):
Dude, this is awesome! You fixed all the issues we had with Option in the first place. I love that you got the compiler to check `#[ts(optional)]` with a great error message

gustavo-shigueo:
> There is a weird note attached to the error though, which I don't know how to get rid of:
> 
> ```
> note: required by a bound in `<OptionalStruct as TS>::inline::check_that_field_is_option`
>    --> ts-rs/tests/integration/optional_field.rs:94:10
>     |
> 94  | #[derive(TS)]
>     |          ^^ required by this bound in `check_that_field_is_option`
> ...
> 104 |     #[ts(optional = nullable)]
>     |     - required by a bound in this function
>     = note: this error originates in the derive macro `TS` (in Nightly builds, run with -Z macro-backtrace for more info)
> ```

With how good the error message is overall, I don't think this is a problem

PR #371, opened 2024-11-25
Fix empty struct tagged enum variant

### LINKED ISSUE #369: bug: Regression Empty Struct Enum Variants with serde tag attribute
**Describe the bug**
Attempting to derive TS for an enum containing empty structs as well as the serde tag attribute fails.

This bug is a regression from v9.1

**Expected behavior**
The code should compile

**Version**
v10.0

**Additional context**
Not Applicable

gustavo-shigueo:
## Goal

Fix incorrect behavior of the `tag` attribute for structs without any fields declared with braces (`struct Foo {}`). These structs, when used with the `tag` attribute should compile and produce a type containing a single field representing the struct's tag.

This bug was revealed by a regression caused #344, which refactored internally tagged enums to take advantage of struct tagging instead of manually reimplementing the behavior.
Closes #369 

## Changes

How did you go about solving the problem?

## Checklist

### DIFF
--- a/macros/src/types/mod.rs
+++ b/macros/src/types/mod.rs
@@ -37,7 +37,7 @@ fn type_def(attr: &StructAttr, ident: &str, fields: &Fields) -> Result<DerivedTS
 
     match fields {
         Fields::Named(named) => match named.named.len() {
-            0 => unit::empty_object(attr, &name),
+            0 if attr.tag.is_none() => unit::empty_object(attr, &name),
             _ => named::named(attr, &name, named),
         },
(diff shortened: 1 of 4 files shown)


gustavo-shigueo:
Hey @NyxCode, since the release of v10, we have fixed a couple of bugs related to serde_json null/boolean and now this one. Support for tokio types has also been added. I think it might be a good idea to release a version 10.1.0 to include these bug fixes

NyxCode:
Good idea!

PR #385, opened 2025-03-06
Fix `#[ts(as = "...")]` dependency list

### LINKED ISSUE #384: bug: Neither #[ts(type = "..")] nor #[ts(as = "..")] imports dependencies
**Describe the bug**
When I override the generated type with either `#[ts(type = "..")]` or `#[ts(as = "..")]`, it generates the expected code, __except__ for the import declarations at the top of the .ts file.

Example:

Attempt 1 (default):

json_map.rs:
```rs
use serde_json::Value as JsonValue;

/// A NewType wrapper around a JSON object (key-value map).
/// Guarantees the inner value is the [`JsonValue::Object`] variant.
#[derive(Debug, Default, Clone, ts_rs::TS, Hash, PartialEq, Eq)]
#[ts(export)]
pub struct JsonMap(JsonValue);
```

JsonMap.ts:
```ts
import type { JsonValue } from "./serde_json/JsonValue";

/**
 * A NewType wrapper around a JSON object (key-value map).
 * Guarantees the inner value is the [`JsonValue::Object`] variant.
 */
export type JsonMap = JsonValue;
```

OK, but this isn't what I want. `JsonMap` should not be any `JsonValue`; it should only ever be a `{ [key in string]?: JsonValue }`.

Attempt 2:

json_map.rs:
```rs
#[derive(Debug, Default, Clone, ts_rs::TS, Hash, PartialEq, Eq)]
#[ts(export)]
#[ts(as = "HashMap::<String, JsonValue>")]
pub struct JsonMap(JsonValue);
```

JsonMap.ts:
```ts

/**
 * A NewType wrapper around a JSON object (key-value map).
 * Guarantees the inner value is the [`JsonValue::Object`] variant.
 */
export type JsonMap = { [key in string]?: number | string | boolean | Array<JsonValue> | { [key in string]?: JsonValue } | null };
```

Closer to what I want but it's basically redefining `JsonValue` when it does not need to, it should just import it. And:
ERROR: `Cannot find name 'JsonValue'.ts(2304)`

Attempt 3:

json_map.rs
```rs
#[derive(Debug, Default, Clone, ts_rs::TS, Hash, PartialEq, Eq)]
#[ts(export)]
#[ts(type = "{ [key in string]?: JsonValue }")]
pub struct JsonMap(JsonValue);
```

JsonMap.ts:
```ts

/**
 * A NewType wrapper around a JSON object (key-value map).
 * Guarantees the inner value is the [`JsonValue::Object`] variant.
 */
export type JsonMap = { [key in string]?: JsonValue };
```

Basically exactly what I want, but I had to quote it literally (why didn't `Map::<String, JsonValue>` yield this?), and again:
ERROR: `Cannot find name 'JsonValue'.ts(2304)`

Attempt 4:

```rs
#[derive(Debug, Default, Clone, ts_rs::TS, Hash, PartialEq, Eq)]
#[ts(export)]
#[ts(type = "{ [key in string]?: any }")]
pub struct JsonMap(JsonValue);
```

Compiles and generates, but again, this is not correct, because it should not be `any`, it should be a valid JSON object value such as is generated for `JsonValue` (`serde_json::Value`).

**To Reproduce**
1. Try the above code in order.
2. Notice the generated .ts files, except for case 1, do not import any dependencies.

**Expected behavior**
I expect the same import to occur when I use `#[ts(as = "...")]`, as the macro should have access to the same type info as if declared that way.

It would also be nice to have a `#[ts(require("MyType", "AnotherType", "JsonValue"))]`, or maybe better: `#[ts(import(...))]`, etc. if using `#[ts(type = "MyType | AnotherType | JsonValue")]` or similar and it cannot infer the required imports.

**Screenshots**

**Version**
10.1.0

gustavo-shigueo:
## Goal

Using `#[ts(as = "...")]` at the top level of a struct would not correctly add the imports to the generated file
Closes #384

## Changes

Add `type_as` to `dependencies`

## Checklist

### DIFF
--- a/macros/src/types/type_as.rs
+++ b/macros/src/types/type_as.rs
@@ -10,12 +10,15 @@ use crate::{
 pub(crate) fn type_as_struct(attr: &StructAttr, name: &str, type_as: &Type) -> Result<DerivedTS> {
     let crate_rename = attr.crate_rename();
 
+    let mut dependencies = Dependencies::new(crate_rename.clone());
+    dependencies.append_from(type_as);
+
     Ok(DerivedTS {
-        crate_rename: crate_rename.clone(),
+        crate_rename,
         inline: quote!(#type_as::inline()),
         inline_flattened: None,
         docs: attr.docs.clone(),
-        dependencies: Dependencies::new(crate_rename),
+        dependencies,
         export: attr.export,
         export_to: attr.export_to.clone(),
         ts_name: name.to_owned(),
@@ -27,12 +30,15 @@ pub(crate) fn type_as_struct(attr: &StructAttr, name: &str, type_as: &Type) -> R
 pub(crate) fn type_as_enum(attr: &EnumAttr, name: &str, type_as: &Type) -> Result<DerivedTS> {
     let crate_rename = attr.crate_rename();
 
+    let mut dependencies = Dependencies::new(crate_rename.clone());
+    dependencies.append_from(type_as);
+
     Ok(DerivedTS {
-        crate_rename: crate_rename.clone(),
+        crate_rename,
         inline: quote!(#type_as::inline()),
         inline_flattened: None,
         docs: attr.docs.clone(),
-        dependencies: Dependencies::new(crate_rename),
+        dependencies,
         export: attr.export,
         export_to: attr.export_to.clone(),
         ts_name: name.to_owned(),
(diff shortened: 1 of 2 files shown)


NyxCode:
Great stuff, thanks for the quick fix!

PR #386, opened 2025-03-06
Prevent naming collisions with <T as crate::TS>

gustavo-shigueo:
## Goal

While writing #385 I noticed the implementation of `TS` for `HashMap` lacked the use of `<T as crate::TS>` in `TS` method calls, which could lead to naming collisions if the user implements some other trait with methods called `name`, `ident` etc, so I took some time to add the trait specifier to every method call I could find. I am not sure I got them all, but I think it should be enough to prevent any issues

### DIFF
--- a/ts-rs/src/lib.rs
+++ b/ts-rs/src/lib.rs
@@ -411,7 +411,7 @@ pub trait TS {
     /// Identifier of this type, excluding generic parameters.
     fn ident() -> String {
         // by default, fall back to `TS::name()`.
-        let name = Self::name();
+        let name = <Self as crate::TS>::name();
 
         match name.find('<') {
             Some(i) => name[..i].to_owned(),
@@ -472,7 +472,7 @@ pub trait TS {
                 }
             }
         }
-        Self::visit_dependencies(&mut Visit(&mut deps));
+        <Self as crate::TS>::visit_dependencies(&mut Visit(&mut deps));
 
         deps
     }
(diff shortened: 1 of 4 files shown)


ts-rs/src/lib.rs
```diff
@@ -411,7 +411,7 @@ pub trait TS {
     /// Identifier of this type, excluding generic parameters.
     fn ident() -> String {
         // by default, fall back to `TS::name()`.
-        let name = Self::name();
```
NyxCode, line 414: I don't think adding the qualifiers was necessary everywhere but within macros. 

For example, here: `Self::name()` can't be anything else other than `<Self as TS>::name()`. This would be ambiguous if and only if the `trait` was defined as `trait TS: SomeOtherTrait { .. }` where `SomeOtherTrait` also had a `name()` method. In that case, we'd also get a compiler error.

Same thing goes for `impl` blocks.  
Remember that, unlike in C++, generics are checked only once, and not for all monomorphizations. If a `trait { .. }` or generic `impl .. { .. }` block compiles, then it works for all monomorphizations.

This is only a problem where we generate a (possibly invalid) `impl` block from a macro.

gustavo-shigueo: Makes sense, the only time I actually got a compiler error about this was the result of the file `macros/src/types/type_as.rs` not using the full qualifier. I just thought "Hm, I'm doing this anyway, might as well do it everywhere" lol

PR #393, opened 2025-03-28
Handle `serde(skip_serializing_if)` + `serde(default)`

### LINKED ISSUE #394: `skip_serializing_if` assumes that object’s field may not exist
`skip_serializing_if` assumes that object’s field may not exist

Serde’s helper-attribute parameter `skip_serializing_if` assumes that object’s field may not be serialized – the field may not exist, so the correct TypeScript signature is `{key?: T}`.

This is true for any value of `skip_serializing_if` parameter and any type of the object’s field. That makes [implementation](https://github.com/Aleph-Alpha/ts-rs/pull/393) quite simple.

Handling this parameter allows to reduce the following boilerplate
```rust
#[derive(Deserialize, Serialize, TS)]
#[ts(export)]
struct Item {
    #[ts(optional)]
    #[serde(skip_serializing_if = "Option::is_none")]
    foo: Option<u8>,
    #[ts(optional)]  // produces compilation error at the moment
    #[serde(skip_serializing_if = "std::ops::Not::not", default)]
    bar: bool,
}
```

To the following code
```rust
#[derive(Deserialize, Serialize, TS)]
#[ts(export)]
struct Item {
    #[serde(skip_serializing_if = "Option::is_none")]
    foo: Option<u8>,
    #[serde(skip_serializing_if = "std::ops::Not::not", default)]
    bar: bool,
}
```

Expected TypeScript output
```typescript
export type Item = { foo?: number, bar?: boolean, };
```

manifest:
## Goal

When both `#[serde(skip_serializing_if)]` and `#[serde(default)]` are present, we know that
- serialization may produce `undefined`
- deserialization accepts `undefined`

This PR adds a special-case for this adding a `?` to the field (like `#[ts(optional = nullable)]` would do).

Closes #394

## Changes

- Refactor handling of `#[ts(optional)]`, moving relevant code into `macros/src/optional.rs`
- Add support for `#[ts(optional)]` on tuple structs
  - We didnt support this before, and nobody complained.
    However, since serde allows `#[serde(skip_serializing_if)]` and `#[serde(default)]` on tuple structs, I felt like our special-case should also work there.
- Make this behavior overridable using `#[ts(optional = false)]`.
  - This is also useful when there is a `#[ts(optional)]` on the struct, allowing it to be overridden for a single field.

## Checklist


gustavo-shigueo:
This is intentionally not supported, see #204

NyxCode:
I have re-opened this for now to let us experiment with the special-cases I outlined in #394 a bit.
The two special-cases I'd like to play with are 

```rust
#[serde(skip_serializing_if = "<anything>", default)]
x: T // => x?: T
```

and

```rust
#[serde(skip_serializing, default)]
x: T // => x?: T
```

ts-rs/tests/integration/serde_skip_serializing.rs
```diff
@@ -0,0 +1,59 @@
...
+    d: Option<i32>,
+}
+
+#[test]
+fn serde_skip_serializing_if() {
+    assert_eq!(
+        Item::decl(),
+        "type Item = { a: number | null, b?: boolean, c?: number | null, d?: number | null, };"
+    );
+}
```
NyxCode, line 59: @manifest I have implemented the two special-cases and have written a lengthy explanation why I believe this is the correct behaviour.

If you find some time, I'd appreciate if you could read the comments, and tell me what you think!

ts-rs/tests/integration/serde_skip_serializing.rs
```diff
@@ -0,0 +1,59 @@
...
+    //      the second wouldn't tell users that `null` is never serialized
+    #[serde(skip_serializing_if = "Option::is_none", default)]
+    c: Option<i32>,
+
+    // Serialization produces:  `         undefined       `
+    // Deserialization accepts: `number | undefined | null`
+    // Most general binding:    `d?: number | null`
+    #[serde(skip_serializing, default)]
+    d: Option<i32>,
+}
```
NyxCode, line 51: @manifest  I have implemented the two special-cases and have written a lengthy explanation why I believe this is the correct behaviour.

If you find some time, I'd appreciate if you could read the comments, and tell me what you think!

gustavo-shigueo (approved):
LGTM

gustavo-shigueo:
Hey @NyxCode, I see you marked the PR as `[WIP]` is there anything else you want to add before merging? If that's because of my commits, I was just resolving the merge conflict

NyxCode:
@gustavo-shigueo Yeah, I think there are some open questions here still - and I couldn't find the "convert to draft" button ^^

gustavo-shigueo:
> Yeah, I think there are some open questions here still - and I couldn't find the "convert to draft" button ^^

Oh okay, lol

NyxCode:
Missing features
- tuple structs. Serde allows skipping serialization for those, so we should handle them as well.
  Next step is probably to support `#[ts(optional)]` for tuple structs first.
- Override, e.g using `#[ts(optional = false)]`

NyxCode:
We could possibly emit an error/warning to prevent users from running into `TS1257: A required element cannot follow an optional element.` for tuple structs.

macros/src/optional.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/// Given a field, return a tuple `(is_optional, type)`.  
+///
+/// `is_optional`:  
+/// An expression evaluating to bool, indicating whether the field should be annotated with `?`.
+///
+/// `type`:  
+/// The transformed type of the field after applying the `#[ts(optional)]` annotation.
+/// For a field of type `Option<T>`, the returned type is `T`. For all other types `T`, it's `T`.
```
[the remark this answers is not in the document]
NyxCode: There's a misunderstanding here. So the function returns a rust `syn::Type`.  

If there is no `#[ts(optional)]` anywhere, you get the same type back you put in.  
For a `#[ts(optional)] field: Option<T>`, you get back `T`.  
For a `#[ts(optional = nullable)] field: Option<T>`, you get back what you put in (`Option<T>`).

PR #398, opened 2025-04-04
Allow arbitrary expressions in top-level `#[ts(rename)]`

NyxCode:
## Goal

Allow arbitrary expressions in `#[ts(rename)]` on structs, enums, and variants.  
This PR does not touch the `#[ts(rename)]` attribute for fields.  

The motivation for this was #397, but I suspect this might be useful in other cases.  
Since this doesn't really complicate the codebase, I thought it might be a nice addition.

## Changes

- In `DerivedTS`, `ts_name` is now of type `Expr`.
- Slight adjustment in code for parsing attributes.
- Instead of the various `name: &str` parameters, we now pass the `Expr` around instead
- Code-generation was slightly adjusted accordingly
- Updated docs & changelog

This is a technically breaking change, since `TS::output_path` now returns `Option<PathBuf>` instead of `Option<&'static Path>`.  
All existing `#[ts(rename = "literal")]` attributes work the same as before though, since `"literal"` is obviously a valid expression.

## Checklist


macros/src/types/enum.rs
```diff
@@ -81,17 +82,17 @@ fn format_variant(
     }
 
     let untagged_variant = variant_attr.untagged;
-    let name = match (variant_attr.rename.clone(), &enum_attr.rename_all) {
+    let ts_name = match (variant_attr.rename.clone(), &enum_attr.rename_all) {
         (Some(rn), _) => rn,
-        (None, None) => variant.ident.to_string(),
-        (None, Some(rn)) => rn.apply(&variant.ident.to_string()),
+        (None, None) => make_string_literal(&variant.ident.to_string(), variant.ident.span()),
+        (None, Some(rn)) => make_string_literal(&rn.apply(&variant.ident.to_string()), variant.ident.span()),
```
NyxCode, line 88: Since the rest of the code now expects an `Expr`, there are a couple of places where we construct an `Expr` from a string

NyxCode:
`cargo +nightly fmt` seemed to have corrected some messed-up some line endings. You'll want to look at the diff with whitespace hidden.

gustavo-shigueo (approved):

gustavo-shigueo:
This is awesome! Can't wait to see where else this can be used

PR #399, opened 2025-04-04
Allow arbitrary expressions in `#[ts(export_to)]`

NyxCode:
This is basically @NimonSour's #347 including tests and implemented like #398.  
Thanks @NimonSour for the idea & initial implementation!

Docs & Changelog are updated, export-logic remains the same, and unlike #398, everything is 100% backwards-compatible.

Note that this implementation only allows for expressions of whose type implements `Display`.  
That excludes `Path` and `PathBuf`. Some more work would need to be done to accept those as well.  
If that's needed in the meantime, calling `.display()` on a `Path`/`PathBuf` makes them `Display`, fixing the issue.


gustavo-shigueo (approved):
LGTM!

PR #400, opened 2025-04-05
CLI

### LINKED ISSUE #54: How to handle naming conflicts
Currently, having multiple types with the same name is problematic.  
Firstly, with just `#[ts(export)]`, they will just overwrite eachother. This can be worked around with `#[ts(export_to = "..")]`.  
Then, when a third type depends on two types with the same name, the generated imports for the third type will be invalid. This can be worked around with `#[ts(rename = "..")]`.  

How can we improve on this? Resolving these conflicts would require coordination between different invocations of `derive(TS)`.  
I don't like to reintroduce something like `export!` to which you pass each type you wish to export, since this gets really painfull when working with a deep module tree, requiring you to make everything public.

### LINKED ISSUE #133: Add index.ts with export for all bindings.

### LINKED ISSUE #408: Feature request: export all types from a single entry point
first of all, thank you for this library.

currently when you want to use the generated types, you have to write import statements for each type individually, unless the export output is one single giant file which is not great.

if the library could generate export statements inside a single entry point(e.x index.ts) you would then be able to write one import statement and include multiple types from there:

`import {Foo,Bar} from 'mylib'`

instead of:

```
import Foo from 'mylib/foo.ts'
import Bar from 'mylib/somewhere/bar.ts'
```

**Additional context**
i actually wrote an implementation for this, but the problem i faced was that the export_all_into function is being called from multiple places at the same time, thus making duplicate export statements in one file, im not familiar with the library internals so if you could guid me to do this it would be great, thank you.

gustavo-shigueo:
I'm copying the original `cli` PR into a new one for a couple of reasons:
- The thread on the old one is VERY long and is getting hard to follow
- Whoever is using the account that opened that PR (if anyone) probably doesn't wanna receive notifications about it

Closes #133
Closes #54
Closes #408

## Todo
- [x] Find a way to only run `ts-rs` tests through this CLI
    -  Done by gating the generation of tests
        - Added a feature flag called `export` to `ts-rs` that re-enables test generation.
        - This feature is always enabled when using this CLI
- [x] Explore a way of generating an `index.ts` file to reexport all the generated types
    - During export, generate a metadata file to list all types and where they are exported to
        - In the CLI, use the metadata file to create the index.ts file
        - Gate the generation of this file behind a feature  flag named `generate-metadata`
- [x] A configuration file so the user doesn't have to pass the flags every time
    - [x] Add `config` flag to pass the path to the file?
    - [x] Add `init` subcommand to help create the config file
- [ ] Discuss other features of the CLI
- [ ] When we consider it ready, it would be interesting to also make it available through `cargo binstall`

## Feature wishlist
- [x] Customize representations (e.g `number` vs `bigint`) (@NyxCode) (being worked on in #334)
- [x] Load configuration from configfile (@NyxCode) (being worked on in #334)
- [x] (maybe) Combine multiple types in one file? (@NyxCode) (#316)
- [x] (maybe) generate `index.ts` file, re-exporting types (@NyxCode) (See explanation above)

## Checklist


cli/src/args.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    /// Generates an index.ts file in your --output-directory that re-exports all
+    /// types generated by ts-rs
+    #[arg(long = "index")]
+    pub generate_index_ts: bool,
+
+    /// Generates only a single index.ts file in your --output-directory that
+    /// contains all exported types
+    #[arg(long = "merge")]
+    pub merge_files: bool,
```
gustavo-shigueo, line 34: I believe that the combination of #316 and #399 has made this flag obsolete, as the user can create a `const` and use it on all `export_to` attributes to get the same effect

gustavo-shigueo:
The implementation of the CLI is a breaking change for anyone who uses `--no-default-features` to disable `serde-compat` as that will also disable the new `export` feature, preventing the test functions that export the types from being generated

gustavo-shigueo:
Hey @NyxCode, I merged #334 into the CLI and added a command to generate a default config file. I also removed the `--merge` flag, as I believe that #316 already does that well enough, especially when combined with #399.

What do you think of the current state of the CLI? Are there other features you want to include?

gustavo-shigueo:
Hey @NyxCode! It looks llike I no longer have write access to ts-rs. Did something happen with you or is this because the repo is in the Aleph Alpha org?

NyxCode:
@gustavo-shigueo Very weird - certainly nothing I did. I'll figure it out asap, sorry about that.

gustavo-shigueo:
No problem. I imagine someone at the org level did some kind of security audit and went "who tf is this random dude with access to one of our repos" lol

cli/src/config.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    fn merge(self, other: Self) -> Self {
+        Self {
+            output_directory: self.output_directory.or(other.output_directory),
+            overrides: other.overrides,
+            no_warnings: self.no_warnings || other.no_warnings,
+            esm_imports: self.esm_imports || other.esm_imports,
+            format: self.format || other.format,
+            generate_index_ts: self.generate_index_ts || other.generate_index_ts,
+            no_capture: self.no_capture || other.no_capture,
```
gustavo-shigueo, line 128: I'm thinking this isn't proper behavior for merging booleans when it comes to CLI vs Config file. I think the CLI should override the config file, but if the config file has  `no_warnings=true` and the CLI has `--no-warnings=false` the config file would win due to the OR operator. Maybe the flags should be `Option<bool>` so the merging would look like `self.no_warnings.or(other.no_warnings)`, then, instead of `if args.no_warnings { /*...*/ }`, we'd have `if args.no_warnings == Some(true) { /*...*/ }` (or `is_some_and(|x| x)` or `unwrap_or(false)`, etc)

CriusNyx:
Is there some work that I could contribute to this to help this PR get merged? I'm needing a way to generate barrel files (those index.ts files that reexport everything).

PS: NyxCode. Nice name. :3

gustavo-shigueo:
Hello @CriusNyx , I'm not quite sure if/when this might be merged, as there are a few possible concerns about it:

- @NyxCode has raised some concerns in #382
- This PR is a breaking change for anyone using `--no-default-features` to disable `serde-compat`
- We are not yet sure if there are more features we want to add
- We haven't documented or tested the CLI very thoroughly yet

All that being said, you can install it with the following command:
`cargo install --git https://github.com/Aleph-Alpha/ts-rs --branch feat/cli ts-rs-cli` and then use a `ts-rs export --index` to achieve the behavior you need

gustavo-shigueo:
Hey @NyxCode, what do you think about merging this into the `next` branch (or `main` after we merge `next`)? We just had someone make another PR (#459) about not wanting `cargo test` to emit the types unless explicitly asked to and their implementation was pretty similar to what this PR is doing (feature flag, enabled by default which, when disabled, does not emit the types. Then running `cargo test export_bindings_` with the feature flag) with major difference being that the `cargo test` command would still be written manually

PR #404, opened 2025-05-11
Remove duplicate validation

### LINKED ISSUE #403: bug: Deriving `TS` should not fail for empty structs in the presence of a `#[serde(rename_all)]` attribute
**Describe the bug**

The implementation of `fn check_attributes(…)` is overly strict, resulting in undesirable compilation failures for totally valid Rust code.

https://github.com/Aleph-Alpha/ts-rs/blob/40b82771c8b63c986d6fcf8a71540d45816418c2/macros/src/types/unit.rs#L64-L74

If at all, then the `TS` derive macro should emit a warning for empty structs with `rename_all` attributes.

Preferably neither an error, nor a warning should be emitted, since there are many valid reasons for adding serde attributes to types, even if they might not have an actual effect in certain scenarios.
As such the type might become empty on certain platforms or only be non-empty given a certain feature. The current overzealous behavior of `ts_rs::TS` prohibits those valid use-cases.

**To Reproduce**

Given the following code, which compiles just fine …

```rust
use serde::Serialize;

#[derive(Serialize)]
#[serde(rename_all = "camelCase")]
pub struct EmptyDto {}
```

Add a `#[derive(ts_rs::TS)]` to `EmptyDto` and the code (despite still being 100% valid) will suddenly fail to compile with the following error:

```rust
error: `rename_all` is not applicable to unit structs
```

**Expected behavior**

… the code should still compile fine, since it is still valid.

**To Fix**

Just delete the following offending lines from `fn check_attributes(…)`:

```rust
if attr.rename_all.is_some() { 
    syn_err!("`rename_all` is not applicable to unit structs"); 
}
```

**Screenshots**

n/a

**Version**

10.1

**Additional context**

n/a

gustavo-shigueo:
## Goal

What is this PR attempting to achieve? Is it a bug fix? Is it related to an issue?
Closes #403

## Changes

Removed the `check_attributes` function from the `unit.rs` file. Attribute validation is already done by the `Attr` trait, in the `assert_validity`. The `check_attributes` function was not only redundant, but also prohibited the use of `#[ts(rename_all = "...")]` in structs defined with braces, but without fields, such as `struct Foo {}`, which should accept `rename_all` even though it would have no effect

## Checklist


gustavo-shigueo:
Thanks! Can we merge it in?

PR #413, opened 2025-06-03
release v11.0.0

NyxCode:


gustavo-shigueo:
Hey @NyxCode, it's been a while since #366, so I may be mistaken, but isn't `#[ts(optional_fields)]` a no-op here, since the fields are not `Option<String>`
image

PR #418, opened 2025-06-10
Add exporting to native typescript enums with string values

xazukx:
## Goal

Add native typescript enums with string values. Integer values are not supported.
I have been using this in a very large codebase with many enum types (simple enums, tuple children, struct children, etc) for a few months with no problems. I've decided to open this PR and bring it to the world.

## Changes

I added the attribute `#[ts(use_ts_enum)]` in order to turn a simple rust enum into a typescript enum. Case renaming still works except kebab renaming, because the name of the enum variant must be a valid typescript identifier. Proper warnings and errors are issued if invalid attributes are used alongside `use_ts_enum` on the same enum.
Inlining an enum in a struct switches back to using union types.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ export type User = { user_id: number, first_name: string, last_name: string, };
 
 ### Features
 - generate type declarations from rust structs
-- generate union declarations from rust enums
+- generate union declarations (and native typescript enums) from rust enums
 - inline types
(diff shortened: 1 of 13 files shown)


stunningpixels:
This is much needed for many production use-cases. Would be great to see this merged.

PR #423, opened 2025-07-07
Replace `import-esm` with `TS_RS_IMPORT_EXTENSION`

gustavo-shigueo:
## Goal

Replace the `import-esm` feature flag with an environment variable called `TS_RS_IMPORT_EXTENSION` as suggested in https://github.com/Aleph-Alpha/ts-rs/discussions/420#discussioncomment-13610108

## Changes

Remove `import-esm` and added `TS_RS_IMPORT_EXTENSION` environment variable

## Checklist


eythaann:
Hey, I don't want to bother you, but my requirement to use `.ts` to use the lib with Deno has been here since pass year 😂 when this will be merged? grateful that forks exist.

gustavo-shigueo:
Hey @eythaann, merging requires review by @NyxCode, until then you can change `Cargo.toml` to point at this branch instead of crates.io

ts-rs/Cargo.toml
```diff
@@ -34,7 +34,6 @@ semver-impl = ["semver"]
 smol_str-impl = ["smol_str"]
 serde-json-impl = ["serde_json"]
 no-serde-warnings = ["ts-rs-macros/no-serde-warnings"]
-import-esm = []
```
NyxCode, line 37: Could we keep the feature itself, just for backwards compatibility?

NyxCode (dismissed):
Hey! Sorry, I have been somewhat busy & now on holiday. 

This looks good, but maybe we could keep the feature to not have to cut a breaking release?

ts-rs/Cargo.toml
eythaann: Definitely, we comment on a discussion thread that the feature should be there and modify the default value of the env.

NyxCode (approved):
thanks for pushing this forward! ❤️

bjones1:
I like this improvement! However, I can't find any mention of it in the docs. Would you please add a section describing this? It probably goes here: https://docs.rs/ts-rs/latest/ts_rs/trait.TS.html#exporting.

NyxCode:
Hey! That hasn't made it into the latest release, but on master, the available environment variables are listed in the crate's root documentation (and with that in the readme).

PR #431, opened 2025-08-31
Change empty object to emit `Record<symbol, never>`

### LINKED ISSUE #429: bug: Empty struct creates ill-formed typescript type in enum with `#[serde(tag = "type")]`
**Describe the bug**
When exporting an empty rust struct, it creates the type `Record<string, never>`. However, when exporting an enum which contains the empty struct as a variant and uses the proc macro `#[serde(tag = "type")]`, it creates an enum variant that cannot exist because the result is a type intersection between `{ "type": ... }` and `Record<string, never>`. Essentially, the `Record<string, never>` prevents the key `"type"` from ever being populated with a value.

**To Reproduce**
main.rs:
```rs
use serde::{Deserialize, Serialize};
use ts_rs::TS;

#[derive(TS, Serialize, Deserialize)]
#[ts(export)]
pub struct FooStatusRequest {}

#[derive(TS, Serialize, Deserialize)]
#[ts(export)]
pub struct BarStatusRequest {
    pub name: String,
}

#[derive(TS, Serialize, Deserialize)]
#[ts(export)]
#[serde(tag = "type")]
pub enum StatusRequest {
    Foo(FooStatusRequest),
    Bar(BarStatusRequest),
}

fn main() {
    println!("Hello, world!");
}
```

```bash
cargo test
```

***Results (concatenated to one file):***
```ts

export type BarStatusRequest = { name: string, };

export type FooStatusRequest = Record<string, never>;
import type { BarStatusRequest } from "./BarStatusRequest";
import type { FooStatusRequest } from "./FooStatusRequest";

export type StatusRequest = { "type": "Foo" } & FooStatusRequest | { "type": "Bar" } & BarStatusRequest;
```

**Expected behavior**
Either a type for the blank struct should be generated that prevents values from being populated while still functioning properly in the type intersection with `{ "type": ... }` (serde tag = "type" component) (though, I don't believe this is currently possible), **or**, it could simply omit including the blank struct in the type intersection:

Second solution possible output sample (given `main.rs` from earlier):
```ts
export type StatusRequest =
      { "type": "Foo" } // Was `& FooStatusRequest`. Type omitted here
    | { "type": "Bar" } & BarStatusRequest;

```

**Version**
11.0.1

**Additional context**
N/A

gustavo-shigueo:
## Goal

Fix bad TypeScript inference when combining a tagged enum with a unit struct
Closes #429

## Changes

Change the empty struct to emit `Record<symbol, never>` instead of `Record<string, never>`

## Checklist

### DIFF
--- a/macros/src/types/unit.rs
+++ b/macros/src/types/unit.rs
@@ -12,7 +12,7 @@ pub(crate) fn empty_object(attr: &StructAttr, ts_name: Expr) -> DerivedTS {
 
     DerivedTS {
         crate_rename: crate_rename.clone(),
-        inline: quote!("Record<never, never>".to_owned()),
+        inline: quote!("Record<symbol, never>".to_owned()),
         inline_flattened: None,
         docs: attr.docs.clone(),
         dependencies: Dependencies::new(crate_rename),
--- a/ts-rs/tests/integration/docs.rs
+++ b/ts-rs/tests/integration/docs.rs
@@ -220,7 +220,7 @@ fn export_c() {
             " *\n",
             " * Testing\n",
             " */\n",
-            "export type C = Record<never, never>;\n",
+            "export type C = Record<symbol, never>;\n",
         )
     } else {
         concat!(
@@ -231,7 +231,7 @@ fn export_c() {
             " *\n",
             " * Testing\n",
             " */\n",
-            "export type C = Record<never, never>;",
+            "export type C = Record<symbol, never>;",
             "\n",
         )
     };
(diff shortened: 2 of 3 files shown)


fxf8:
Please check https://github.com/Aleph-Alpha/ts-rs/issues/429 once more.

fxf8:
> > Seems reasonable, though I wonder if this is really a breaking change, because `Record<symbol, never>` is more permissive than `Record<string, never>`, but I still agree with your decision merge into a side-branch.
> 
> Even though this change shouldn't affect anyone (other than fixing the bug in #429), it does, technically, change the generated type, which could break something if, for example someone did (for reasons I can't possibly imagine lol) `keyof UnitStruct`, which would give `string`, now gives you `symbol`
> 
> Which reminds me, I forgot to add the breaking change label to this PR lol
> 

I did not think about that. Good for bringing it up.

PR #446, opened 2025-10-14
Make hashmap keys required unless the key is an enum

### LINKED ISSUE #374: Feature request: HashMap with required keys
A way to instruct hashMaps to be generated as records with required keys

### LINKED ISSUE #442: New `HashMap` export type creates type problems
Hello! Long time no see.
I've just upgraded past 10.0.0 with this breaking change:
> Change how HashMap<K, V> is represented in TypeScript. The resulting bindings ({ [key in K]?: V } instead of { [key: K]: V }) are more accurate and flexible.

This does create problems, because now inside a `Object.values(...)`/`Object.entries(...)` the value can be `undefined`. **Which it is definitely not**.

Image

Either using `HashMap<K, V>` or `{ [key: K]: V })` would prevent this problem.

I cannot quite understand why that decision was made.
**But all I ask for is a feature flag to change this please. 🙏🏽**
(for me this has created >25 errors in my repo - all wrongfully, because `undefined` is never the case)

gustavo-shigueo:
## Goal

Remove optional key from hashmaps unless the key type is an enum
Closes #442
Closes #374

## Changes

Added new const to `TS` trait that determines whether or not a type should have a `?` when used as a hashmap key

## Checklist

### DIFF
--- a/ts-rs/src/lib.rs
+++ b/ts-rs/src/lib.rs
@@ -979,6 +979,18 @@ impl<T: TS, const N: usize> TS for [T; N] {
     }
 }
 
+// FIXME: Remove in future release
+// Whether to treat `HashMap` (and derived types) like we did in v11.
+fn use_v11_hashmap() -> bool {
+    static USE_V11_HASHMAP: OnceLock<bool> = OnceLock::new();
+    *USE_V11_HASHMAP.get_or_init(|| {
+        let Ok(var) = std::env::var("TS_RS_USE_V11_HASHMAP") else {
+            return false;
+        };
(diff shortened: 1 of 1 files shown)


timephy:
> Maybe we should make it configurable through environment variables

I propose configuration like this to be done with feature flags 🙏🏽

gustavo-shigueo:
Hey @NyxCode, what would you want this env variable to look like?

By that I mean, do you want a bool that if set to true reverts back to the previous behavior? The other way around? Or something else entirely?

Also, Merry Christmas! Hope you're doing great! Cheers

PR #454, opened 2025-10-25
Fix trait bound generation when using `#[ts(optional)]` on an `Option<Generic>` 

### LINKED ISSUE #453: bug: Auto Genric Bounds failing for `#[ts(optional)]`
**Describe the bug**
Optional generic fields break with `#[ts(optional)]` if they are second generic or don't have implicit TS trait.
Please look at examples, they show it clearly.

**To Reproduce**
```rust
// Will break:
#[derive(Serialize, TS)]
#[ts(export, export_to = "optional_field/")]
struct GenericOptionalStruct<T1, T2>
{
	#[ts(optional)] // <- error is happening here
	a: Option<T2>,
	#[ts(optional = nullable)]
	b: Option<T1>,
	c: Option<T1>,
}

// Works:
#[derive(Serialize, TS)]
#[ts(export, export_to = "optional_field/")]
struct GenericOptionalStruct<T1, T2>
{
	#[ts(optional = nullable)] // <- set to nullable
	a: Option<T2>,
	#[ts(optional = nullable)]
	b: Option<T1>,
	c: Option<T1>,
}

// Works:
#[derive(Serialize, TS)]
#[ts(export, export_to = "optional_field/")]
struct GenericOptionalStruct<T1, T2>
where   // <- added TS requirements
	T1: TS,
	T2: TS,
{
	#[ts(optional)]
	a: Option<T2>,
	#[ts(optional = nullable)]
	b: Option<T1>,
	c: Option<T1>,
}

// Works: 
#[derive(Serialize, TS)]
#[ts(export, export_to = "optional_field/")]
struct GenericOptionalStruct<T> // <- removed T2
where 
	T: TS,
{
	#[ts(optional)]
	a: Option<T>,
	#[ts(optional = nullable)]
	b: Option<T>,
	c: Option<T>,
}
```

**Expected behavior**
For the first  example to work

**Version**
`11.1.0`

**Additional context**
I haven't found out why exactly the issue is happening. may add later.

gustavo-shigueo:
## Goal

Fix dependency tracking for generic type when using `#[ts(optional)]`

Using `#[ts(optional)]` on a field of type `Option<T>`, where `T` is an actual generic type in the struct declaration, caused the generation of the `impl` block to fail to include the `where T: TS` clause. This was caused by the fact that the `used_type_params` function couldn't see through the type generated by the application of `#[ts(optional)]`, which is `<T as ::ts_rs::TS>::OptionInnerType` and would fail to detect the use of the type parameter `T`

To be clearer, the following type

```rs
#[derive(TS)]
struct Foo<T> {
  #[ts(optional)]
  foo: Option<T>
}
```

Would generate the following `impl` block:

```rs
impl<T> TS for Foo<T> {
  // ...
}
```

Instead of the correct `impl` block:

```rs
impl<T> TS for Foo<T> where T: TS {
  // ...
}
```

Closes #453

## Changes

Added a match arm in `used_type_params` that handles the case `Type::Path(TypePath { qself: Some(_), .. })`

I also found a way to simplify the compile-time checking that ensures `#[ts(optional)]` cannot be used in non-option fields, with the added bonus that it moves the compiler error to the field's type instead of the attribute itself and makes it a lot shorter:

Before
image

After
image

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -744,6 +744,35 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d75a2a4b1b190afb6f5425f10f6a8f959d2ea0b9c2b1d79553551850539e4674"
 
 [[package]]
+name = "jiff"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3590fea8e9e22d449600c9bbd481a8163bef223e4ff938e5f55899f8cf1adb93"
+dependencies = [
+ "jiff-tzdb-platform",
+ "log",
+ "portable-atomic",
+ "portable-atomic-util",
+ "serde",
+ "windows-sys",
+]
+
+[[package]]
+name = "jiff-tzdb"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "68971ebff725b9e2ca27a601c5eb38a4c5d64422c4cbab0c535f248087eda5c2"
+
+[[package]]
+name = "jiff-tzdb-platform"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "875a5a69ac2bab1a891711cf5eccbec1ce0341ea805560dcd90b7a2e925132e8"
+dependencies = [
+ "jiff-tzdb",
+]
+
+[[package]]
 name = "js-sys"
 version = "0.3.74"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


ts-rs/tests/integration/optional_field.rs
```diff
@@ -21,6 +21,27 @@ fn in_struct() {
...
+    f: i32,
+}
+
+#[test]
+fn in_generic_struct() {
+    assert_eq!(
+        GenericOptionalStruct::<()>::decl(),
+        "type GenericOptionalStruct<T> = { a?: T, b?: T | null, c: T | null, d: T, e: number | null, f: number, };"
+    )
+}
```
sinder38: the integration test does will work with or without the changes.
Use my example from the issue:
```rust
#[derive(Serialize, TS)]
#[ts(export, export_to = "optional_field/")]
struct GenericOptionalStruct<T1, T2>
{
	#[ts(optional)] // <- error is happening here
	a: Option<T2>,
	#[ts(optional = nullable)]
	b: Option<T1>,
	c: Option<T1>,
}
```

gustavo-shigueo: Ah, you're right, I tried adding more fields after I finished the fix to be thorough but I forgot that the test only fails if the ONLY field that mentions `T` is the one with `#[ts(optional)]`, i.e., lines 31 - 34 ruin this test. I'll remove the extra fields

NyxCode:
Thank you for the investigation & fix! Very subtle issue in `used_type_params`!

macros/src/optional.rs
```diff
@@ -77,17 +77,15 @@ pub fn apply(
...
+            parse_quote!(true),
             if nullable {
                 field_ty.clone()
             } else {
-                unwrap_option(crate_rename, field_ty)
+                // expression that evaluates to the the Option's inner type,
+                // but fails to compile if `field_ty` is not an `Option`.
+                parse_quote_spanned! {
+                    span => <#field_ty as #crate_rename::IsOption>::Inner
+                }
```
NyxCode: This is simpler than the previous check, though I suspect it sacrifices on the diagnostic emitted.

PR #455, opened 2025-11-05
refactor: fix clippy lints

sandr01d:
## Goal

Fixes clippy lints without introducing any actual changes. The lint regarding usage of `parse` instead of `from_str_radix` shows up in projects using `ts-rs` as a dependency due to it being used in the derive macro for `TS`. This trips over the CI in the project I'm using `ts-rs` with.

## Changes

Implements the changes suggested by clippy without introducing any changes in behavior. I added an `#[allow(clippy::enum_variant_names)]` for the `Optional` enum because changing the names of the variants would not make any sense IMO.
There are two lints left regarding `std::path::absolute` only being available in Rust 1.79.0 and the project using 1.78.0. Guess the best way to resolve that is to update the Rust version, so I left out those ones.

## Checklist


gustavo-shigueo:
> There are two lints left regarding `std::path::absolute` only being available in Rust 1.79.0 and the project using 1.78.0.

This had me really concerned that the MSRV check was giving invalid passes until I checked that both uses are in the tests lol

gustavo-shigueo:
LGTM

PR #465, opened 2026-01-15
Added rust_decimal support with support for all serialization formats

Isaac-Duarte:
## Goal

This PR adds support for the [rust_decimal](https://github.com/paupino/rust-decimal) crate which we use at my company for financial calculations.

## Changes

Mirrored the pattern of the `bigdecimal` implementation

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,6 +19,17 @@ checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
 
 [[package]]
 name = "ahash"
+version = "0.7.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


NyxCode:
Hey! This has been attempted in #273 and #351 in the past.  
The TL;DR is that `rust_decimal` is somewhat abusing cargo feature flags, and these past attempts have failed because of it.  
Since then, my stance has been that this would need to be implemented within `rust_decimal`.

However, thinking about this again, there might be a workaround to get this working without support from `rust_decimal`.  
We could write a custom `impl TS for Decimal`, and try to detect which format is used. We'd need to call `Decimal::serialize`, and then try to determine what TS type would fit best.

Isaac-Duarte:
Thanks for providing that context. I would not mind prototyping a solution to make it possible to detect this. Let me know if you would like me to implement something like this!

Thanks!

PR #472, opened 2026-01-31
Refactor: Better handling of environment variables

NyxCode:
This PR is an attempt at cleaning up how binding generation is configured.

Currently, some impls just read environment variables on their own and adjust their behavior.  
This PR changes that. Environment variables are read beforehand when the user (or the generated test) calls `Config::from_env`. That `Config` instance is then passed around to `TS::decl`, `TS::inline`, etc.  

So now, all configuration can be cleanly set programmatically without environment variables.  
I hope this helps users who don't use `#[ts(export)]`, but do that manually.

Also, this would pave the way to experiment with a config file (without necessarily introducing a CLI). The tests we generate for exporting could, instead of `Config::from_env`, call something like `Config::from_file` in the future.  
This is out-of-scope for this PR for now. But even if we don't end up doing that, I still hope this is a good change.

The PR is big, since plumbing the `Config` through everything just does that. Sorry for that.

---

- This PR also removes the deprecated import-esm feature.
- This PR also removes the check that TS_RS_IMPORT_EXTENSION is `js` or `ts`.
  I felt like it wasn't needed, but we can add it back if I'm missing something.

---

## TODO

- [x] review (& possibly adjust) documentation

### DIFF
--- a/ts-rs/tests/integration/arrayvec.rs
+++ b/ts-rs/tests/integration/arrayvec.rs
@@ -1,5 +1,5 @@
 #![cfg(feature = "arrayvec-impl")]
-use ts_rs::TS;
+use ts_rs::{Config, TS};
 
 #[derive(TS)]
 #[ts(export, export_to = "arrayvec/")]
@@ -11,5 +11,6 @@ struct ImStackAllocated {
 
 #[test]
 fn arrayvec() {
-    assert_eq!(ImStackAllocated::decl(), "type ImStackAllocated = { smol_vec: Array<number>, name: string, nested: Array<Array<string>>, };")
+    let cfg = Config::from_env();
+    assert_eq!(ImStackAllocated::decl(&cfg), "type ImStackAllocated = { smol_vec: Array<number>, name: string, nested: Array<Array<string>>, };")
 }


macros/src/types/tuple.rs
```diff
@@ -80,9 +80,9 @@ fn format_field(
         .map(|t| quote!(#t.to_owned()))
         .unwrap_or_else(|| {
             if field_attr.inline {
-                quote!(<#ty as #crate_rename::TS>::inline())
+                quote!(<#ty as #crate_rename::TS>::inline(cfg))
```
NyxCode: Within `macros/`, all `quote!` calls just use `cfg` like this, since it's guaranteed to be in scope (the generated function definition has the `cfg: Config` parameter)

ts-rs/src/lib.rs
```diff
@@ -657,16 +608,95 @@ impl Dependency {
...
+        Self {
+            large_int_type: "bigint".to_owned(),
+            use_v11_hashmap: false,
+            export_dir: "./bindings".into(),
+            import_extension: "".to_owned(),
+        }
+    }
+}
+
+impl Config {
```
NyxCode: I'm not sure about the API of `Config` yet. It's just a plain struct right now, with `new`/`default`, `from_env`, getters (like `out_dir()`) and setters that return `Self` for chaining (like `with_out_dir(dir)`).

ts-rs/src/lib.rs
```diff
@@ -460,23 +440,29 @@ pub trait TS {
     /// placeholders, resulting in a generic typescript definition.
     /// Both `SomeType::<i32>::decl()` and `SomeType::<String>::decl()` will therefore result in
     /// the same TypeScript declaration `type SomeType<A> = ...`.
-    fn decl() -> String;
+    fn decl(cfg: &Config) -> String {
+        panic!("{} cannot be declared", Self::name(cfg))
+    }
```
NyxCode: Before, there was no default implementation for `decl`, `decl_concrete`, and `inline_flattened`.  
Since it seemed like it was implemented to just panic very often, I made this the default implementation.  Can't think of any downside atm.

ts-rs/src/lib.rs
gustavo-shigueo: I think this is fine for now, especially since, AFAIK, there is no invalid combination of properties for `Config` - by this I mean, it's not invalid to have, for example, `use_v11_hasmap: true` and `import_extension: None`.

If this ever happens then it might be a good idea to revisit this api to make sure such invalid constructions cannot be created (maybe with a typestate builder?)

ts-rs/src/lib.rs
```diff
@@ -657,16 +590,147 @@ impl Dependency {
...
+            cfg = cfg.with_out_dir(dir);
+        }
+
+        if let Ok(ext) = std::env::var("TS_RS_IMPORT_EXTENSION") {
+            if !ext.trim().is_empty() {
+                cfg = cfg.with_import_extension(Some(ext));
+            }
+        }
+
+        #[allow(deprecated)]
```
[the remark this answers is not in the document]
gustavo-shigueo: Nevermind, I just saw the `with_v11_hashmap` function lol

gustavo-shigueo (approved):
Awesome! No more functions randomly scattered all over the place to read `std::env::var` lol

PR #474, opened 2026-02-02
Allow `#[ts(flatten)]` for unit structs

### LINKED ISSUE #473: TS Export Panics When Flattening Zero-Field Struct
**Describe the bug**

When using `#[serde(flatten)]` with a zero-field struct that derives `TS`, the TypeScript export generation panics with the error: `"[StructName here] cannot be flattened"`.

**To Reproduce**

```rust
Create a zero-field struct with `TS`:
#[derive(Deserialize, Serialize, TS)]
#[ts(export)]
pub struct Inner {}

// Create a parent struct that flattens the zero-field struct:
#[derive(Deserialize, Serialize, TS)]
#[ts(export)]
pub struct Outer {
    #[serde(flatten)]
    pub inner: Inner,
    pub other_field: String,
}
```

3. Run TypeScript export 
4. See panic: `"Inner cannot be flattened"` 
5. [Surprised pikachu](https://c.tenor.com/UZJd1pjj4NMAAAAd/tenor.gif)

**Expected behavior**

The zero-field struct should be flattened without issue, resulting in TypeScript output equivalent to:
```typescript
export type Outer {
    other_field: string;
}
```

Or at least the library should panic  indicating that zero-field structs cannot be flattened.

**Version**

12.0.1

**Workaround:**

I could just remove the whole struct and it's flatten.

**Why?**

That empty struct is a placeholder which could change at any moment I don't want to remove it even though it is empty

gustavo-shigueo:
## Goal

Allow the use of `#[ts(flatten)]` with unit structs `struct Foo;` and empty structs `struct Foo {}` but not empty tuple structs `struct Foo()`. This matches the behavior of `#[serde(flatten)]`

The use of the type `{ }` when flattening these structs does not cause the `{}` type to appear in typescript because of our `.replace(" } & { ")` which we added back when we first changed the implementation of `flatten` to use type intersections, which means these structs simply disappear from the ts type as though `#[ts(skip)]` was used instead, which also matches serde behavior.

Allowing `flatten` on `struct Foo;` (no curly braces) broke the generation of internally tagged enums because it caused an interaction with the `struct` version of `#[ts(tag = "...")]`, but this ended up making the enum generation code simpler as it removed an unnecessary `match` expression and the need to manually add the internal tag, as the `struct` version of this attribute does it instead

Also added a validation rule to prevent the use of internal tagging (`#[ts(tag = "...")]` without `#[ts(content = "...")]`) when the enum contains tuple variants,
Closes #473

## Changes

How did you go about solving the problem?

## Checklist

### DIFF
--- a/macros/src/attr/enum.rs
+++ b/macros/src/attr/enum.rs
@@ -245,6 +245,15 @@ impl Attr for EnumAttr {
                 item;
                 "content cannot be used without tag"
             ),
+            (false, Some(_), None) => {
+                for variant in item.variants.iter() {
+                    if let Fields::Unnamed(ref unnamed) = variant.fields {
+                        if unnamed.unnamed.len() > 1 {
+                            syn_err_spanned!(variant; r#"`#[ts(tag = "...")]` cannot be used with tuple variants"#);
+                        }
+                    }
+                }
+            }
             _ => (),
         };
 
--- a/macros/src/attr/enum.rs
+++ b/macros/src/attr/enum.rs
@@ -248,6 +248,15 @@ impl Attr for EnumAttr {
                 item;
                 "content cannot be used without tag"
             ),
+            (false, Some(_), None) => {
+                for variant in item.variants.iter() {
+                    if let Fields::Unnamed(ref unnamed) = variant.fields {
+                        if unnamed.unnamed.len() > 1 {
+                            syn_err_spanned!(variant; r#"`#[ts(tag = "...")]` cannot be used with tuple variants"#);
+                        }
+                    }
+                }
+            }
             _ => (),
         };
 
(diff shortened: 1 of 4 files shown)


NyxCode (approved):
Code-wise this looks good to me. Fixing a bug while simplifying the code is just great.  
If you'd like me to review the semantics of this change, then I'll need to put aside some time to look into this more. It's been a while since I worked on that part of the code.

However, if you're confident, you can also merge this as-is for a patch release. Your call!

gustavo-shigueo:
Hey @NyxCode, sorry I took a while to respond.

> If you'd like me to review the semantics of this change, then I'll need to put aside some time to look into this more

While I *think* this shouldn't break anything, I do think it's worth having you take a look just to make sure :D

> However, if you're confident

Nah, I'm **never** confident in my own code lmao

PR #477, opened 2026-02-08
Allow flattening newtypes and as types

cairomassimo:
## Goal

1. Allow flattening newtypes if the inner type can be flattened. This matches `serde_json` behavior.
2. Allow flattening types with `#[ts(as = "...")]` if the referenced type can be flattened. This can be useful and (as far as I can tell) has no downsides.

## Changes

- `macros/src/types/newtype.rs`: Add `inline_flattened` that delegates to the inner type
- `macros/src/types/type_as.rs`: Add `inline_flattened` that delegates to the `as` type
- `macros/src/types/enum.rs`: Restructure internally tagged enum handling to check for newtype variants first, ensuring they always use intersection syntax (`{ "tag": "A" } & Type`) rather than the flattened path
- Added tests in `newtype_flatten.rs` and `type_as_flatten.rs`

## Checklist


gustavo-shigueo:
Hi @cairomassimo thanks for the PR! We are already making a change (#474) to how we flatten structs with zero fields that touches some of the same areas of the code as you have here - specifically, the match expression in the enum file. Would you mind waiting until those changes are merged as to help us avoid a possible conflict and breaking change?

cairomassimo:
Sure! The changes to the implementation are small anyway, and the test
won't have any conflicts. Thanks for the quick response

PR #485, opened 2026-03-07
Add support for representing enums as "const object"s

jthacker:
## Goal

This PR introduces a new representation for enums: `#[ts(repr(enum = const_object))]`.

While ts-rs already supports exporting enums as TypeScript enums or type unions, there are many cases where a TypeScript const object is preferred. This approach provides:

* **Runtime access**: The values are available as a real object at runtime (unlike type unions).
* **Type safety**: A companion type alias is generated using keyof typeof, ensuring the type stays in sync with the object.
* **Flexibility**: It supports both integer discriminants and string-based unit variants.

This is a common syntax used in modern TypeScript.
Additionally, as stated in the [TypeScript docs](https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums) the "biggest argument in favour of this format over TypeScript’s enum is that it keeps your codebase aligned with the state of JavaScript."

### Macros
- Added `Repr::ConstObject` to the `Repr` enum to support the new attribute.
- Updated the attribute parser to recognize `const_object` in `#[ts(repr(enum = ...))]`.
- Implemented the TypeScript code generation for the `const_object` representation in both `inline()` and `decl()`. This includes logic to handle implicit and explicit integer discriminants.
- Updated variant formatting to correctly support the `const_object` representation, ensuring it falls back to string names for unit variants without discriminants.
 
### Documentation
- Added documentation for the new `const_object` representation, outlining its usage and requirements.

### Testing
- Added comprehensive tests covering:
    - Enums with explicit and implicit integer discriminants.
    - String-based unit variants.
    - Integration with `rename_all` strategies (snake_case, camelCase, kebab-case).
    - Correct inlining behavior.

### Other
- Ran `cargo +nightly fmt` as required by contributing guide resulting in many changes unrelated to this change so probably a gap in CI

## Checklist

- [X] If necessary, I have added documentation related to the changes made.
- [X] I have added or updated the tests related to the changes made.

### DIFF
--- a/macros/src/attr/enum.rs
+++ b/macros/src/attr/enum.rs
@@ -229,7 +229,7 @@ impl Attr for EnumAttr {
             }
 
             if let Optional::Optional { .. } = self.optional_fields {
-                syn_err!("`optional_fields` is not compatible with `as`");
+                syn_err!("`optional_fields` is not compatible with `repr`");
             }
         }
 
--- a/macros/src/attr/enum.rs
+++ b/macros/src/attr/enum.rs
@@ -41,10 +41,11 @@ pub enum Tagged<'a> {
     Untagged,
 }
 
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, PartialEq, Eq)]
 pub enum Repr {
     Int,
     Name,
+    ConstObject,
 }
 
 impl EnumAttr {
--- a/macros/src/attr/mod.rs
+++ b/macros/src/attr/mod.rs
@@ -241,6 +241,7 @@ fn parse_repr(input: ParseStream) -> Result<Repr> {
     let ident = content.parse::<Ident>()?;
     match ident.to_string().as_str() {
         "name" => Ok(Repr::Name),
-        _ => syn_err!(span; "expected `name`"),
+        "const_object" => Ok(Repr::ConstObject),
+        _ => syn_err!(span; "expected `name` or `const_object`"),
     }
 }
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -294,6 +294,36 @@ impl DerivedTS {
 
                 buffer.trim_end_matches(['|', ' ']).into()
             },
+            Some(Repr::ConstObject) => quote! {
+                let variants = #inline;
+                let mut variants = variants.split(',').map(|v| v.trim()).peekable();
+
+                if variants.peek().is_none() {
+                    return "never".into()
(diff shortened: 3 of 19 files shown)


gustavo-shigueo:
Hey @jthacker, thanks for the PR, I have a couple of questions:
- Have you checked that this works properly with export merging (multiple types exported to the same file)?
- I imagine such enums would be forbidden from having any kind of complex variant such as `Foo { x: i32 }` or `Foo(String)`, has this  been added to the attribute validation?
- Will the fact that the object and the type have the same name cause issues when importing it?

jthacker:
@gustavo-shigueo Thanks for the feedback! I've answered inline.

> Have you checked that this works properly with export merging (multiple types exported to the same file)?

I wasn't aware of this feature, something I'm interested in using too. I have extended the `same_file_export.rs` integration test in the [latest commit](https://github.com/Aleph-Alpha/ts-rs/pull/485/changes/f9dd66ec80c200c7115cd6645ab9dfa19f67f70d). However, these tests appear to not verify the output (just that generation doesn't fail), so let me know if you'd like a more robust test. I manually verified (visually and with tsc) and it looks as I would expect.

> I imagine such enums would be forbidden from having any kind of complex variant such as `Foo { x: i32 }` or `Foo(String)`, has this  been added to the attribute validation?

Yes, this is forbidden. This style only works with unit variants. The rust compiler will fail with an error like this `All variants of an enum marked as #[ts(repr(enum))] must be unit variants`. This is utilizing the existing [code path for enum repr](https://github.com/Aleph-Alpha/ts-rs/blob/7182ad8289596097235406b715fa04506443b4ad/macros/src/attr/enum.rs#L226). I documented this [here](https://github.com/Aleph-Alpha/ts-rs/pull/485/changes#diff-7962a4b0d656b7969ae8d7a967fa9c774f84ac70c76383a3c2ca10cbcbede4ecR350).

> Will the fact that the object and the type have the same name cause issues when importing it?

It works as expected. If you only want the type, you can import it with `import type MyType from ...`. Otherwise, importing with `import MyType from ...` gives you access to both. I believe TypeScript keeps these names in separate namespaces and can use the appropriate one as it sees fit.

jthacker:
@gustavo-shigueo a friendly ping, anything else you'd like to see in this pr?

gustavo-shigueo:
Hey, not particularly, just need @NyxCode to take a look

PR #486, opened 2026-03-07
Fix export order inconsistency

### LINKED ISSUE #479: Race condition when generate multiple type into same file
Hi, I'm using ts-rs to generate multiple enum into same ts file, and i use `git diff` after `cargo test` to make sure all bindings are up-to-date.
But I notice my CI failed sometime due to the order of the types changed in the generated file.

For example both `Foo` and `Bar` has `#[ts(export, export_to = "Output.ts")]`. And in `Output.ts`, its 
```ts
export enum Foo { }

export enum Bar { }
```
Sometimes it became Bar first 
```ts
export enum Bar { }

export enum Foo { }
```

I wonder is this expected and unavoidable? Maybe I should just generate them in seperate file instead?

Thanks in advance!

gustavo-shigueo:
## Goal

Fix inconsistent export order
Closes #479

## Changes

Change merge code to no longer depend on the `export type` prefix

## Checklist


B-2U:
It looks good to me! Though, in my humble opinion, the `.nth(2)` deserves a comment.

B-2U:
It seems that Nyx has been inactive for months :(

gustavo-shigueo:
He has mentioned before that he's quite busy at work, I'm sure when he has some time he'll take a look at the PRs but it's not easy making time for open source when your boss is constantly breathing down your neck lol

macros/src/lib.rs
```diff
@@ -1,4 +1,3 @@
-#![macro_use]
```
gustavo-shigueo, line 1: Apparently newer versions of the compiler hate this

NyxCode (approved):
Looks like a nice improvement! 
(I'm still not 100% comfortable with that approach, still being somewhat fragile, but I can't come up with something more robust that doesn't introduce a lot of bloat (like extra metadata comments or a full TS parser))

gustavo-shigueo:
Hey @NyxCode, I see you've approved the PR but the merge button is still locked
image

NyxCode:
@gustavo-shigueo Seems like something changed in the org, adding a default ruleset I cant get rid of. I changed your role to admin now, so you should be able to bypass it. Feel free to check if you can remove the rule, but i dont think we can.
If we can't get rid of the org-wide rule, it might make sense to sign commits in the future to get rid of the warning - although I don't think it matters for us.

gustavo-shigueo:
I see, I never even looked into signing commits lol, I'll check it out and make sure to use it for future prs

PR #487, opened 2026-03-09
Allow customizing the export note

jancespivo:
## Goal

Allow customizing the export note. My use case is sending to LLM. There the note is irrelevant and also can occur additional costs. So I do `Config::new().with_custom_export_note("")` 

The first iteration was simple boolean flag `omit_export_note` however this is more generic. I can switch back to the boolean flag (the setter will be `without_export_note()`).

## Changes

Adds an `export_note` field to `Config` and a setter method to replace the default file generation comment.

## Checklist


jancespivo:
To make it more robust I can add something like `format!("// {}\n\n", note)` to the constructor, though `// ` should be added to every line to not break comment and keep it valid ts.

gustavo-shigueo:
Some of the export merging code is quite sensitive to changes to the start of the file, including the note. I don't really think this is worth the potential problems and future maintenance burden. I'll close this for now until/unless @NyxCode decides he wants to add this

PR #492, opened 2026-05-04
Implementing astrolabe compatibility

aurorameccanicadev:
## Goal

Implements a new `astrolabe-impl` feature for compatibility with [astrolabe](https://docs.rs/astrolabe/latest/astrolabe/) type `Date`, `DateTime` and `Time`.

## Changes

I mimicked the implementations for other crates. I added an integrations test guarded by the `astrolabe-impl` feature.

My only doubt is about auto-format. I tried running it locally, but it formats many other files I haven't edited. Should I apply the autoformat to the whole crate or only to the files I modified/created?

## Checklist

- [X] If necessary, I have added documentation related to the changes made.
- [X] I have added or updated the tests related to the changes made.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -75,6 +75,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "astrolabe"
+version = "0.5.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cce4de37eb70265f1fe35d8d39a63d3b7ff1a71284dbb2d1fd72ca5766e7409f"
+
+[[package]]
 name = "autocfg"
 version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1498,6 +1504,7 @@ name = "ts-rs"
 version = "12.0.1"
 dependencies = [
  "arrayvec",
+ "astrolabe",
  "bigdecimal",
  "bson",
  "bytes",
--- a/README.md
+++ b/README.md
@@ -125,6 +125,7 @@ We are currently waiting for [#54140](https://github.com/rust-lang/rust/issues/5
 | tokio-impl         | Implement `TS` for types from *tokio*                                                                                                               |
 | jiff-impl          | Implement `TS` for types from *jiff*                                                                                                                |
 | arrayvec-impl      | Implement `TS` for types from *arrayvec*                                                                                                            |
+| astrolabe-impl     | Implement `TS` for types from *astrolabe*
 
 ### Contributing
 Contributions are always welcome!
--- a/ts-rs/Cargo.toml
+++ b/ts-rs/Cargo.toml
@@ -38,6 +38,7 @@ no-serde-warnings = ["ts-rs-macros/no-serde-warnings"]
 tokio-impl = ["tokio"]
 jiff-impl = ["jiff"]
 arrayvec-impl = ["arrayvec"]
+astrolabe-impl = ["dep:astrolabe"]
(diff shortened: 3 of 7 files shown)


aurorameccanicadev:
I was reading more into the astrolabe library and ts-rs. My only doubt is how to correctly implement compatibility with types [Precision](https://docs.rs/astrolabe/latest/astrolabe/enum.Precision.html) and [Offset](https://docs.rs/astrolabe/latest/astrolabe/enum.Offset.html), which are enums.

The only other missing type is [CronSchedule](https://docs.rs/astrolabe/latest/astrolabe/struct.CronSchedule.html), but I don't think it makes sense to implement it, as it doesn't even have a Display/ToString method, its goal is only to build iterators yelding DateTimes.

gustavo-shigueo:
Hey @aurorameccanicadev, thanks for the PR!

> My only doubt is about auto-format. I tried running it locally, but it formats many other files I haven't edited. Should I apply the autoformat to the whole crate or only to the files I modified/created?

Yeah, we've been slacking a bit on the auto-format recently, you can leave it alone. I'll open a PR just for that at some point

> I was reading more into the astrolabe library and ts-rs. My only doubt is how to correctly implement compatibility with types [Precision](https://docs.rs/astrolabe/latest/astrolabe/enum.Precision.html) and [Offset](https://docs.rs/astrolabe/latest/astrolabe/enum.Offset.html), which are enums.

Yeah, those are a bit more complicated. Since it's not possiblie to use `#[derive(TS)]` on them and none of our declarative macros like `impl_primitives!` are intended for enums, the only real way to do it would be to manually implement the TS trait for them.

From what I've seen of them, `Precision` might be a case where we'd want to implement it in a way that mimics `#[ts(repr(enum))]`, while `Offset` could be implemented in a way that mimics `Option<i32>`.

This is mostly a guess based on their definitions though. I've taken a look at their serde implementation and these types have no serde support, so I don't know what the best TS representation for them would be or if you need one at all

gustavo-shigueo:
The following code could be used as a starting point. It is based on my guesses about them described above. Again, it might be the case that you don't need to support these at all

Be aware that I'm typing this directly as a GitHub comment, so I have NOT run this through the compiler (i.e. it might have compiler errors) nor have I tested it

```rs
impl TS for Offset {
    type WithoutGenerics = Self;
    type OptionInnerType = Self;

    fn name(cfg: &Config) -> String {
        "Offset".to_owned()
    }

    fn inline(cfg: &Config) -> String {
        "number | null".to_owned()
    }
}

impl TS for Precision {
    type WithoutGenerics = Self;
    type OptionInnerType = Self;

    fn name(cfg: &Config) -> String {
      "Precision".to_owned()
    }

    fn inline(cfg: &Config) -> String {
        "0 | 2 | 3 } 6 } 9".to_owned()
    }
    
    fn decl(cfg: &Config) -> String {
        format!("enum {} {{ Seconds = 0, Centis = 2, Millis = 3, Micros = 6, Nanos = 9 }}", Self::name())
    }
}
```

aurorameccanicadev:
Agree, I guess exporting these additional types to typescript is not useful so far, as they're not serde-serializable.

Also I noticed that DateTime objects are serialized to RFC339 format ( [ref](https://github.com/GiyoMoon/astrolabe/blob/88d1e27386b063e4e302d4ca25b7cb0a177af15d/src/serde/datetime.rs#L13) ), which includes the timezone offset already.

gustavo-shigueo (approved):

PR #494, opened 2026-06-16
Support `heapless::String`

mmalecki:
## Goal

Type generation for `heapless::String`, unrelated to an existing issue.

## Changes

* Add an `impl_shadow` for `heapless::String`
* Fix shadowing of `String` in macros
* Add a test

## Checklist

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -72,7 +72,7 @@ impl DerivedTS {
         let docs = match &*self.docs {
             [] => None,
             docs => Some(quote! {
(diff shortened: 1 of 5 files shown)


gustavo-shigueo:
Hi @mmalecki, thanks for the PR! Wow, somehow it never even ocurred to me that the return types could also get shadowed, I pretty much only ever worried about method calls such as `<T as ::ts_rs::TS>::inline()` lol

gustavo-shigueo (approved):

PR #497, opened 2026-07-21
Adds support for `nonempty` and `either` crates

Niedzwiedzw:
## Goal

Adds support for external crates:
 - https://github.com/cloudhead/nonempty
 - https://github.com/rayon-rs/either

## Changes

Followed implementation of serde_json compatibility module to implement support for this crate.

## Checklist


Niedzwiedzw:
@NyxCode sorry for ping but could you take a look? I think its a useful change

ts-rs/Cargo.toml
```diff
@@ -67,3 +71,5 @@ tokio = { version = "1", features = ["sync"], optional = true }
 jiff = { version = "0.2", optional = true }
 arrayvec = { version = ">= 0.6, < 0.8", optional = true }
 astrolabe = { version = "0.5.4", optional = true }
+nonempty = { version = "0", optional = true }
```
NyxCode, line 74: I think this would accept any "0." version, which might introduce breaking changes.

ts-rs/Cargo.toml
```diff
@@ -9,16 +9,16 @@ homepage = "https://github.com/Aleph-Alpha/ts-rs"
...
+  "development-tools::build-utils",
+  "wasm",
+  "web-programming",
 ]
 readme = "../README.md"
 rust-version = "1.78.0"
 
 [features]
-default = ["serde-compat"]
+default = ["serde-compat", "either-impl"]
```
NyxCode, line 21: This should not be enabled by default

NyxCode (changes requested):
Thanks for the PR. Two issues, see my comments above. Have you checked the generated TS bindings against the serde_json output of the two crates?

ts-rs/Cargo.toml
Niedzwiedzw: I don't think this crate really would, api is really simple - also this allows user to stick to an older version? I will pin to latest though

Niedzwiedzw:
> Thanks for the PR. Two issues, see my comments above. Have you checked the generated TS bindings against the serde_json output of the two crates?

yes yes I did. thanks for review -  I applied the requested changes cc @NyxCode 

EDIT: feel free to squash the commits, or I can do it if you want

PR #499, opened 2026-09-03
Support panicking on unknown serde attributes, with new feature `serde-errors`

### LINKED ISSUE #467: Feature request: Panic On Unknown Serde Attribute
Currently when an unknown serde attribute is encountered a warning is emitted. There is also `no-serde-warnings` which suppresses the warning, but really what I want to happen is that it fails to compile to tell me about this issue.

There should be a new flag `panic-on-unknown-serde` (or `strict-serde`) where it panics when an attribute is unknown. This can save hours of debugging.

joshtriplett:
Fixes: https://github.com/Aleph-Alpha/ts-rs/issues/467


joshtriplett:
Checking back in about this.

